PR #45, opened 2022-10-10
Added Go beavs

jpmorgan98:
Alterations:
- Added a testing directory in /mcdc/
- started some unit test utilities
- altered the read me


ilhamv:
you accidentally added the cache you silly goose

jpmorgan98:
actually real commit that we need to commit:
- added testing directory
- some updates to  the readme (added some banners)
- added something to the gitignor

PR #54, opened 2022-10-12
input_ unit tests for material initialization

aaronjamesreynolds:
Wrote tests to verify material card created from input arguments with:
* single neutron energy group and no dnp group
* single neutron energy group and single dnp group
* single neutron energy group and multiple dnp groups
* multiple neutron energy groups and no dnp group
* multiple neutron energy groups and single dnp group
* multiple neutron energy groups and multiple dnp groups

These tests specify all optional input parameters when initializing a material. Tests still need to be written to ensure the expected default values are assigned when optional parameters are not specified.

There's also a minor bug fix: nu_s was being indexed as a 2D array when it is 1D.


ilhamv (approved):
Looks good to me!

PR #55, opened 2022-10-13
Basic PRNG Unit Test

spasmann:
What I did:
Created a new file `test_kernel.py`, with a new unit test `test_rn_basic()` which verifies 10 random numbers with reference data from the Forrest Brown. Pytest passes all tests.

Future
We should consider adding a dummy material to the default input_card.


ilhamv (approved):
Looking good! This tests not only `rng` but also `rng_skip_ahead*`.

PR #64, opened 2022-12-21
Changed array_equal to isclose in regression tests

spasmann:
Seems different machines were generating slightly different results in the regression testing causing them to fail. Regression tests pass after changing np.array_equal to np.isclose.


ilhamv (approved):
I just learned that we can do `np.allclose(a,b)` instead of `np.isclose(a,b).all()`. But they are equivalent.

PR #65, opened 2023-02-03
iQMC Fixed-Source Algorithm

spasmann:
iQMC is functioning in MCDC for fixed source, multigroup, problems. Tested with Reeds multimedia problem and a high-density-polyethylene multigroup problem in infinite medium. iQMC _should_ work on 2D/3D problems but this has yet to be tested.  The Halton and Sobol Sequences are the available LDS along with ''random''. 

To run any problem with iQMC the ''material index' and ''fixed source'' variables must be defined and passed through.  iQMC uses a separate tally for the scalar flux, so no need to use MCDC's traditional tally. The iQMC mesh and tallies are saved in the same H5 file as traditional grids/tallies.

Currently, the Source Iteration (equivalent to a Picard Iteration) is the only available iterative method. Krylov methods including GMRES and LGMRES will follow, as will k-eigenvalue problems.

All tests are passing.


spasmann:
Should also mention that all tests are passing.

mcdc/card.py
```diff
@@ -116,6 +116,31 @@ def reset(self):
...
+                          'iqmc_fixed_source': np.ones([1, 1, 1, 1]),
+                          'iqmc_material_idx': np.ones([1, 1, 1, 1]),
+                          'iqmc_flux': np.ones([1, 1, 1, 1]),
+                          'iqmc_mesh': {'x': np.array([-INF, INF]),
+                                        'y': np.array([-INF, INF]),
+                                        'z': np.array([-INF, INF]),
+                                        't': np.array([-INF, INF]),
+                                        'mu': np.array([-1.0, 1.0]),
+                                        'azi': np.array([-PI, PI]), },
+                          'sensitivity': 0,
```
ilhamv, line 142: Redundant?

ilhamv (approved):

PR #77, opened 2023-02-08
code in black

ilhamv:


jpmorgan98 (approved):
changes did not alter function, test suite passed

PR #81, opened 2023-02-13
Cooper2D iQMC input deck and some functionality changes

spasmann:
- The weight rouletting function is now toggleable similar to implicit capture
- If not in iQMC mode, the iQMC structs are initialized with size zero
- Added input deck for Cooper's 2D fixed source problem
- Regression test for cooper iqmc problem
- Black styling

### DIFF
--- a/mcdc/type_.py
+++ b/mcdc/type_.py
@@ -517,7 +517,7 @@ def make_type_technique(card):
     Ng = card.materials[0]["G"]
     N_dim = 6  # group, x, y, z, mu, phi
     if not card.technique["iQMC"]:
-        Nx = Ny = Nt = Nmu = N_azi = N_particle = Ng = N_dim = 0
+        Nx = Ny = Nz = Nt = Nmu = N_azi = N_particle = Ng = N_dim = 0
 
     struct += [("iqmc_mesh", mesh)]
     # Low-discprenecy sequence


mcdc/type_.py
```diff
@@ -504,20 +504,23 @@ def make_type_technique(card):
...
-    mesh, Nx, Ny, Nz, Nt, Nmu, N_azi = make_type_mesh(card.technique["iqmc_mesh"])
-    struct += [("iqmc_mesh", mesh)]
 
-    # Low-discprenecy sequence
+    mesh, Nx, Ny, Nz, Nt, Nmu, N_azi = make_type_mesh(card.technique["iqmc_mesh"])
     N_particle = card.setting["N_particle"]
     Ng = card.materials[0]["G"]
     N_dim = 6  # group, x, y, z, mu, phi
+    if not card.technique["iQMC"]:
+        Nx = Ny = Nt = Nmu = N_azi = N_particle = Ng = N_dim = 0
```
ilhamv, line 520: Should we add `Nz` too?

spasmann:
```suggestion
        Nx = Ny = Nz = Nt = Nmu = N_azi = N_particle = Ng = N_dim = 0
```

PR #92, opened 2023-03-10
iQMC Eigenmode

spasmann:
Added Power Iteration capabilities to iQMC for eigenvalue problems. Also added two example and regression test problems.


mcdc/card.py
```diff
@@ -125,9 +125,13 @@ def reset(self):
             "wr_target": 1.0,
             "iQMC": False,
             "iqmc_generator": "sobol",
+            "fixed_source_solver": "source_iteration",
```
ilhamv, line 128: May need to give the 'iqmc' prefix to the new attribute names.

mcdc/card.py
```diff
@@ -125,9 +125,13 @@ def reset(self):
...
+            "eigenmode_solver": "power_iteration",
```
ilhamv, line 129: Same here

mcdc/main.py
```diff
@@ -38,7 +38,12 @@ def run():
     # Run simulation
     # TODO: add if iQMC execute iqmc_loop()
```
ilhamv, line 39: Delete

ilhamv (approved):

PR #100, opened 2023-04-24
Updated Regression/Verification Testing

bwhewe-13:
Updated the regression and verification testing process. Relocated the regression/ input files to a problems/ folder separated by fixed_source and eigenvalue. 

In the regression tests, tests were added for running problems in both python and numba modes and comparing the results between MPI and numba.
Pytest markers were used to distinguish the difference between different tests.

With verification, the input files are read from the problems/ folder. The verify.py file allows for running the verification problems after a single failed regression test, as seen in the conftest.py file. No work was done on the process.py and plot.py files for the verification problems.


ilhamv:
Hi Ben.
This looks great! But there are things that we need to discuss, as I think they still need to be improved to give the best experience to the developers setting up new tests and running tests. Let's resolve them one by one. First, instead of having `test_serial` and `test_processor` can we handle it (serial and parallel) as a parameter? For example, besides `mode` we will also have parameter `N_mpi`, and both are passed into a single test function `test`, where inside it we will check whether N_mpi is one or larger than one. What do you think?

bwhewe-13:
Hi Ilham @ilhamv, 
Yes, the number of processors can be used a parameter if there are a set number of processors you would like to use (1, 2, 4, 8) or if you would like it to be an option for the user, I could set it up as an argument when running pytest from the command line.

ilhamv:
Yes. And yes, it would be great if we could make it an option for the user, with the default `N_mpi=1` or a serial run.

Next, for the `particles` option, I think it would be best if we keep it fixed and defined (by the test writers) in the problem definition because some problem runs much longer than others, and only the test writer would know how many particles should be run. (There should be some criteria on the test documentation, though, like the minimum number of particles and/or maximum runtime in serial, as general guidance for the test writer). What do you think?

bwhewe-13:
I was adding the number of particles as an option to be able to reuse the code for the verification tests, but I can make them unchangeable.

ilhamv:
> I was adding the number of particles as an option to be able to reuse the code for the verification tests, but I can make them unchangeable.

Ah, I see. That makes sense. How about instead, we make `N_particle` an option inside the MC/DC code, which overrides the `N_particle` set in the setting card? The point is that we are trying to minimize the number of lines needed to be written/copied in setting up each test function. What do you think?

bwhewe-13:
Okay, that makes sense to me. I will work on that.

ilhamv:
Effort is continued in #131

PR #114, opened 2023-08-23
Local rng

braxtoncuneo:
This pull request adds per-particle prng, as well as a number of refactors to support reproducibile prng in highly parallel contexts.

To summarize:

- Instead of having just `rng`, we now have `stateful_rng` and `stateless_rng`
 - `stateful_rng` accepts an object with a `rng_seed` field and performs the usual LCNG iteration upon it, updating the field with the new seed state and returning the value normalized (that is - with uniform distribution between 0.0 and 1.0)
 - `stateless_rng` behaves similarly accept it simply receives a `numba.uint64` instead of an object, using that value as the seed state and not storing the subsequent seed state anywhere
- Instead of skipping ahead through the rng sequence of the LCNG, "child" seed states are produced through hashing. This is done both when initializing the seed state of particles "children" as well as when generating per-iteration seed states in loops. This is done because being able to seek forward and backward is not required by these subsequent seed states. In truth, all that is needed is a reproducible, parallelizable source of seeds, and hashing avoids some potential pitfalls of skip-ahead-based child-seeding. For example, in cases where multiple loops are nested and prng states are required at each layer, some prng states may inadvertently be duplicated due to the offsets of different loop layers aligning with one-another. Specifically, for two nested loops with offset coefficients *X* and *Y*, we can expect duplicate seeds for nested loop iterations (i_1,j_1) and (i_2,j_2) where (i_1*X+j_1*Y) == (i_2*X+j_2*Y). This arises due to the commutativity/associativity of these offsets. Hashing does not have this issue.
- The prng functions now rely upon seed states and all prng constants being type `numba.uint64`. This is important for ensuring reproducibility. A stray mixing of signs or the introduction of a floating point number can make the prng both less random and also make it not align between numba/python mode runs.
- Regression test answers have been updated
- Bugs in the regression tests, unit tests, and verification tests have been fixed

### DIFF
--- a/mcdc/kernel.py
+++ b/mcdc/kernel.py
@@ -84,23 +84,26 @@ def sample_discrete(group, P):
 
 
 @njit(numba.uint64(numba.uint64, numba.uint64))
-def wrapping_mul(a,b):
+def wrapping_mul(a, b):
     return a * b
+
 
 @njit(numba.uint64(numba.uint64, numba.uint64))
 def wrapping_add(a, b):
     return a + b
 
-def wrapping_mul_python(a,b):
+
+def wrapping_mul_python(a, b):
     a = numba.uint64(a)
     b = numba.uint64(b)
-    with np.errstate(all='ignore'):
+    with np.errstate(all="ignore"):
         return a * b
+
 
 def wrapping_add_python(a, b):
     a = numba.uint64(a)
     b = numba.uint64(b)
-    with np.errstate(all='ignore'):
+    with np.errstate(all="ignore"):
         return a + b
 
 
@@ -109,8 +112,6 @@ def adapt_rng(object_mode=False):
     if object_mode:
         wrapping_add = wrapping_add_python
         wrapping_mul = wrapping_mul_python
-
-
 
 
 @njit(numba.uint64(numba.uint64, numba.uint64))
@@ -166,7 +167,7 @@ def rng_skip_ahead_(n, mcdc):
 
 @njit(numba.uint64(numba.uint64))
 def rng_(seed):
-    return wrapping_add(wrapping_mul(RNG_G,seed), RNG_C) & RNG_MOD_MASK
+    return wrapping_add(wrapping_mul(RNG_G, seed), RNG_C) & RNG_MOD_MASK
 
 
 @njit


mcdc/loop.py
```diff
@@ -41,11 +47,11 @@ def loop_main(mcdc):
                 kernel.tally_closeout_history(mcdc)
 
             # Print progress
-            with objmode():
-                print_progress_eigenvalue(mcdc)
+            # with objmode():
```
ilhamv, line 50: Why is this commented-out?

mcdc/loop.py
```diff
@@ -24,13 +27,16 @@
...
     while not simulation_end:
+        cycle_seed = kernel.int_hash_combo(loop_index, mcdc["rng_seed"])
         # Loop over source particles
-        loop_source(mcdc)
+        loop_source(cycle_seed, mcdc)
 
         # Loop over source precursors
         if mcdc["bank_precursor"]["size"] > 0:
-            loop_source_precursor(mcdc)
+            loop_source_precursor(cycle_seed, mcdc)
```
ilhamv, line 39: We are going to need to carefully treat this one.

mcdc/loop.py
```diff
@@ -41,11 +47,11 @@ def loop_main(mcdc):
...
 
             # Print progress
-            with objmode():
-                print_progress_eigenvalue(mcdc)
+            # with objmode():
+            # print_progress_eigenvalue(mcdc)
 
             # Manage particle banks
-            kernel.manage_particle_banks(mcdc)
+            kernel.manage_particle_banks(cycle_seed, mcdc)
```
ilhamv, line 54: May ned to carefully treat this one too.

ilhamv:
The test is failing because of the overflow warning, I guess?

mcdc/kernel.py
```diff
@@ -97,39 +98,135 @@ def rng_skip_ahead(n, mcdc):
...
+def stateful_rng(state, mcdc):
+    seed = state["rng_seed"]
+    g = mcdc["setting"]["rng_g"]
+    c = mcdc["setting"]["rng_c"]
+    mod = mcdc["setting"]["rng_mod"]
+    mod_mask = mod - numba.uint64(1)
+
+    state["rng_seed"] = (g * seed + c) & mod_mask
+    result = state["rng_seed"] / mod
+    return result
```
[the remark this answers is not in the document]
braxtoncuneo: Agreed. This latest round of commits performs this refactor.

mcdc/loop.py
braxtoncuneo: Each argument now gets its own, unique hash. Convert the hex to ASCII for a fun surprise.

mcdc/loop.py
braxtoncuneo: Each argument now gets its own, unique hash. Convert the seed hex to ASCII for a fun surprise.

braxtoncuneo:
No rush. Enjoy the weekend: :smile:

ilhamv:
@braxtoncuneo , I made some changes:
1. Removing redundant unit casting, since `key` and `seed` are cast in `murmur_hash64a` anyway.
2. Renaming `int_hash*` into `split_seed`, which I think would be more descriptive.
3. Merging source sampling into `source_particle`, which I think makes the code more streamlined. This allows us to do the next modification.
4. Using particle seed instead of the `sampling_seed` to sample the source. This last modification however would require us to reset the regression test answers.

Let me know what you think.

ilhamv:
I'm wondering if it is possible to anticipate the overflow warning in Python mode...

braxtoncuneo:
Thanks for the help! I agree with your changes.

This latest set of changes should fix the overflow warnings. The uint64 casts in the python versions of the addition/multiplication function should guarantee the input and output types. Normally, I would just jit the function with an explicit type signature to have these guarantees, but overflow warnings appear to be reported even from jitted functions. This appears to be the case so long as the program is not being run in nopython mode. The `with np.errstate` statements suppress all errors/warnings that are encountered in the body, but they only work outside of nopython mode.

To accommodate both modes, I went with an adapter-based solution. If you think it wise, I can move the adapter code to a separate module. With harmonize integration coming up, it may make sense to have a module designated for mode-dependent or target-dependent logic.

ilhamv (approved):

PR #117, opened 2023-09-05
move out iqmc sweep counter

ilhamv:


spasmann:
This works for source iteration, but we also need to add `mcdc["technique"]["iqmc_sweep_counter"] += 1` to the linear operator kernel functions. Not sure how to make the changes on the PR myself.

ilhamv:
Ah I didn't know about that. I added the counter into the functions.

spasmann (approved):

PR #119, opened 2023-09-08
Parallelization of LDS and Additional Unit Tests

spasmann:
In iQMC the entire LDS used to be stored on each processor. Now, the sequence is split according to the processor rank.

In the unit tests, I added a function called `iqmc_dummy_mcdc_variable()`, the idea is we can call this function, or a similar one, this and it returns the global `mcdc ` (variable generated after `prepare()` in `main.py`). The current input parameters are based on the kornreich eigenvalue problem, but this could be changed to any input deck - maybe we have multiple.

This avoids the issue of having to define the global variable manually in each test. Looking for feedback on this, I think it's very useful, but implementation may not be perfect. Like maybe we want to store` iqmc_dummy_mcdc_variable()` and other dummy input decks in a separate file or something.


ilhamv:
I love the dummy mcdc variable idea!

ilhamv (approved):

PR #129, opened 2023-11-09
Adding domain decomposition 

shac170:
Basic implementation of domain decomposition, no memory reduction yet.


spasmann:
I wanted to update this PR because there had been some changes to `main` since it was opened. I had to merge with my fork's main though which is why the first three commits are the same from my iqmc [pr/140](https://github.com/CEMeNT-PSAAP/MCDC/pull/140).

After that, there were just a few changes that needed to be made to get DD running in numba mode again.  I also added two regression tests. Currently configured, the tests only run if 4 MPI processes are present but we can change this to whatever we'd like, it just needs to be a specific number.

DD is reproducing results nicely on fixed source problems where there is only _one processor per domain_. If there is more than one per domain or the problem is k-eigenvalue I have not been able to reproduce results. @shac170 mentioned this is not expected behavior and is working to figure out the issues.

PR #142, opened 2023-12-20
Add continuous energy physics

ilhamv:


jpmorgan98:
Here is a bash script to configure a cross section library. It sets `MCDC_XSLIB` for both the current enviroment and stores it to `~/.bashrc`. It can be ran with `bash config_xsec_lib.sh`. When this PR is approved I can also add it as a callable option from the `install.sh` script.

*Dependencies:* wget, tar, bash (same as the install script)

File name: `config_xsec_lib.sh`
```
#!/bin/bash

# downloading library from zenodo
wget https://zenodo.org/records/10426266/files/mcdc_xs.tar.gz

# untaring file
tar -xvzf mcdc_xs.tar.gz

# deleting the orignal tar ball
rm mcdc_xs.tar.gz

# going into the cross section library directory
cd mcdc_xs

# getting the present working direcotry
xsec_dir=$(pwd)

# exporting the needed enviroment variable for current enviroment
export MCDC_XSLIB="${xsec_dir}"

# sotring that enviroment variable in bashrc for future runs
export MCDC_XSLIB="${xsec_dir}">> ~/.bashrc 

# printing a complete message
echo "MCDC_XSLIB set as $MCDC_XSLIB in ~/.bashrc"
```

jpmorgan98:
# Update
I updated the install script to pull from a private github page within the CEMeNT org. All members hsould have access to it but it does require that you have set up sshkeys. 

If this is implemented then these two lines should be added to the `.gitignore` to avoid an accidental commit of the unzipped files
```
# Material Data
mcdc_xs
MCDC-Xsec
```

Issue #126, opened 2023-09-29
Updating Regression Tests and Examples


jpmorgan98:
The MPI issues keep failing out after repos are forked and we need to do better about testing all functionality. The specific tests we need to add:
- [x] Parallel regression test
- [x] Numba regression test
- [x] MPI & Numba Regression tests
- [x] Updating examples to have current functionality
I think as a stop gap measure we can write a number of SLURM or LSF batch scripts for Lassen or Quartz to at least somewhat optimize the process tho let me know what yall think.

ilhamv:
Effort to address this: #131

jpmorgan98:
I think we should retool the tests to not be so long. I will tackle this!

ilhamv:
Also, note that now we do mpi test with mpiexec=2 and mpiexec=4. We can just do one of them.

jpmorgan98:
Ya I think a github action only gives us 2 CPU cores anyways.

spasmann:
Dang. 4 would be ideal for DD tests.

jpmorgan98:
We can still launch 4 threads in that case they just wont be at all performant

Issue #138, opened 2023-12-08
Visualization merging


jpmorgan98:
I am trying to merge the visualizer into main but am having some difficulties

jpmorgan98:
# Background
@RohanPankaj added a visualization capabilities as a REU project over summer 2023. I am trying to merge it and recon with the alterations in MC/DC and with some of the new dependency.

In the current visualization is handled via [Netgen](https://ngsolve.org/) a full fat multiphyscis FEA solver that has x86 builds in [conda](https://anaconda.org/conda-forge/netgen). The problem is that Netgen currently requires a [manual build on ARM MacOS](https://docu.ngsolve.org/latest/install/installmacnative.html) and has no listed support for ppc64 (Power9). We don't need visualization as a strict requirement for MC/DC so I might make the configuration and testing of Netgen a non-standard option requiring additional flags to our current [install script](https://github.com/CEMeNT-PSAAP/MCDC/blob/main/install.sh)

## Next Steps 
- [ ] Handle merge between new MC/DC and old MC/DC
- [ ] Configure Netgen on my local machine validate visualizer with new MC/DC version
- [ ] Understanding Netgen as a dependency and install requirements

spasmann:
Curious do we know if the other dependencies,`ngsolve `and `distinctipy`, have similar issues?

jpmorgan98:
The dependencies have all been added and seem to be working on MAC and windows. I had to do some bug hunting to make sure all tests pass but as of PR #143 things seem to be working. When that PR is approved I will close this issue

jpmorgan98:
ngsolve is a pip only command for osz-arm64 (M1/2/3)

PR #109, opened 2023-07-25
Add model visualization

RohanPankaj:

PR #145, opened 2024-01-05
Testing Refactor, New Documentaiton, & Cont Energy Xsec Config script

jpmorgan98:
This probably should have been multiple PRs but oh well.

# New features:
- Testing refactor: Adds osx64 testing target (in-conjunction with linux64), splits regression test into multiple runners to make it go faster! See issue #126 *all the same tests are running*
- Adding some new documentation to generally describe the contribution process to MC/DC
- Continuous energy config: Runs a separate bash script that can be individually launched. To run must be part of the cement center

There are no new dependence, physics, or CS back ends

### DIFF
--- a/.github/workflows/mpi_numba_reg.yml
+++ b/.github/workflows/mpi_numba_reg.yml
@@ -1,10 +1,14 @@
-name: test
+name: Regression Test - Numba and MPI
 
 on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+        fail-fast: false
+        matrix:
+            os: ["ubuntu-latest", "macos-latest"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up python 3.11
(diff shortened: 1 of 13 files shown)


clemekay:
I cloned your branch and tried to install and get the following error:

> PackagesNotFoundError: The following packages are not available from current channels:
>   - ngsolve
>   - distinctipy
>
> Current channels:
> 
>   - https://repo.anaconda.com/pkgs/main/osx-arm64
>   - https://repo.anaconda.com/pkgs/main/noarch
>   - https://repo.anaconda.com/pkgs/r/osx-arm64
>   - https://repo.anaconda.com/pkgs/r/noarch

Do those need to be installed from specific channels?

clemekay:
Update: distinctipy needed the condaforge channel. For ngsolve, their available installations are osx-64, linux-64, and win-64; I've got an Apple M2 and would need osx-arm64 or noarch :/

jpmorgan98:
So this new dependincy was actually added back in PR #109. Apple Sillicon M1 issues where known and discussed in issue #138. We actually don't need the parts of ngsolve that are incompatible so it still works. It just requires a pip install. I will change this in the install script.

clemekay:
Gotcha, didn't think to check the visualizer thread. I also haven't been able to build ngsolve from their source instructions or forums, so I'm glad that's not necessary!

clemekay (approved):
Just commenting this here because I didn't realize we'd already known that ngsolve and distinctipy don't work via conda installation for Apple M1. The way the
`conda install package1 package2 ...`
command works, if any of the packages can't be found it voids the whole command and nothing installs. So having any packages in that line that end up being unavailable voids the whole install script!

PR #146, opened 2024-01-12
Visualization Timescale and color fix.

RohanPankaj:
Added a tick_interval field to the visualize() function to allow users to have more control over the time scale. This is particularly useful when the end time is less than one. A spinbox was also added to the timescale to make adjustment easier. 

Now, users are also able to specify the color of materials by using the material_colors field in visualize(). To do this, pass in a dictionary of the form {"material name":RGB} where RGB values are a list.

### DIFF
--- a/examples/godiva-mockup-visulization/input.py
+++ b/examples/godiva-mockup-visulization/input.py
@@ -73,7 +73,9 @@ mcdc.cell(
 
 mcdc.source(x=[-22.0, 22.0], time=[0.0, 5.0], isotropic=True)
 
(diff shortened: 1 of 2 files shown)


ilhamv:
Hi @RohanPankaj, the black-style test failed. To make it pass, you can install `black` and then apply it to the script that you modify.

RohanPankaj:
Hey @ilhamv just applied it. See "969072e13417b94c2fcf204a2205681d3aa8a1a3"

ilhamv (approved):

PR #150, opened 2024-01-24
Fixing capabilities on continuous energy mode

ilhamv:


jpmorgan98:
Do we have any verification/regression tests for continuous energy?

ilhamv:
Good point. But no, not yet. How should we deal with the data library in the GitHub workflow?

jpmorgan98 (approved):
These all look good and classify as bug fixes. Eventually we will need documentation and testing but we have to wait on those while we get info from the higher ups.

jpmorgan98:
Pull in updates then merge the request!

jpmorgan98:
Just to put this here--we are getting guidance from PSAAP on how best to do continuous energy transport with export control considerations

PR #154, opened 2024-02-06
Keep collision-inducing neutron alive when applicable. Update on tests.

ilhamv:


ilhamv:
Note that performance tests are moved to a separate repo.

jpmorgan98 (approved):

PR #162, opened 2024-02-21
Minor bug in install.sh 

ilhamv:
At some point, we accidentally removed `install` from `conda install mpi4py`.
Is it possible to test `install.sh` as part of the github workflow?


jpmorgan98:
Yes that's a great idea! Could we move this pr to the dev branch?

PR #165, opened 2024-02-27
Add python version check to install script.

clemekay:
In what is now line 27 (was line 18), the word 'install' was removed in commit 650d73c; I added that back. I also added a version check. 

We have the pip installation which is cleaner and definitely suggested for general use. The installation script is useful when trying to debug source code, so I'd still like to keep it up-to-date.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Check python version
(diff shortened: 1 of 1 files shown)


jpmorgan98:
Changing the base from main to dev

PR #188, opened 2024-04-27
Removing upper limit of python version in pyproject

clemekay:


jpmorgan98 (approved):
Removing upper limit simple change. I will merge upon test completion

PR #192, opened 2024-04-30
Update Read the Docs refs

### LINKED ISSUE #191: Setup.py file has mcdc version as 0.1.0
We have a setup.py file that has mcdc's version as 0.1.0. Is this setup file still used for anything? It hasn't been updated in 6 months, so I'd assume not.

clemekay:
Also resolves #191.


jpmorgan98 (approved):
great!

PR #194, opened 2024-04-30
Domain Decomposition Revamped

braxtoncuneo:
Revised domain decomposition functionality to:

- use non-pickled API (`Isends`/`Recv` in place of `isend`/`recv`), which should allow for more particles per send/receive transaction. This refactor required:
  - Adding two additional types to `type_`, which are MPI representations of `type_.particle_record` and of a new "turnstile event", which is used to coordinate halting. This is necessary, since the non-pickled API can only handle custom types when given MPI representations of those types.
  - Adding a global variable *outside of object mode*, called `requests`, which contains a list of request/buffer pairs. This is necessary because `objmode` frees the memory associated with all local variables, regardless of outstanding references, and asynchronous sends need their corresponding request object and message buffer to persist until they are complete. Since `objmode` also disallows direct global variable access, requests and buffers are passed to a full Python (non-object mode) function, which then handles storage of outstanding requests and cleanup of resolved requests.
- coordinate/detect halting conditions without the use of collective operations, which should allow ranks to exchange work without requiring synchronization with all other ranks. This is achieved through the addition of two new message types (in addition to the messages that simply transfer particles - which now has `tag=1`).
  - `dd_turnstile_event` messages (`tag=2`), which communicate a `busy_delta` and `send_delta`, which respectively represent changes to the number of ranks that are occupied with work and the number of particles that have been sent but not received. This message type is only sent to rank zero, which acts as custodian to `busy_total` and `send_total` values, representing the counts affected by these deltas. Once both of these counts reach zero as part of the same transaction, rank zero notifies all other ranks that the system has reached a halting condition. Rank zero does not send this message type to itself, instead directly updating the count when necessary.
  - `halt` messages (`tag=3`), which communicate that a halting condition has been reached, and which is sent out  by rank zero to all other ranks once the aforementioned halting condition has been reached.
- use an all-purpose `dd_distribute_bank` function, which de-duplicates the bank-sending logic performed along each cardinal direction.
- use `dd_check_in` and `dd_check_out` functions, that set up and close out the global state required to coordinate halting and clean up outstanding request objects.
- store numba-mode-compatible domain-decomposition-related information in a newly created `domain_decomp` field of the `type_.global_` type.

### DIFF
--- a/mcdc/input_.py
+++ b/mcdc/input_.py
@@ -1417,9 +1417,9 @@ def domain_decomposition(
     z : array_like[float], optional
         Location of subdomain boundaries in z (default None).
     exchange_rate : float, optional
-        number of particles to acumulate in the domain banks before sending.
+        Number of particles to acumulate in the domain banks before sending.
     work_ratio : array_like[integer], optional
-        Inte
+        Number of processors in each domain
 
     Returns
     -------
--- a/mcdc/input_.py
+++ b/mcdc/input_.py
@@ -1401,7 +1401,7 @@ def domain_decomposition(
     x=None,
     y=None,
     z=None,
-    exchange_rate=100,
+    exchange_rate=100000,
     work_ratio=None,
     repro=True,
 ):
--- a/mcdc/input_.py
+++ b/mcdc/input_.py
@@ -1428,7 +1428,6 @@ def domain_decomposition(
     """
     card = mcdc.input_deck.technique
     card["domain_decomposition"] = True
-    card["domain_bank_size"] = int(1e5)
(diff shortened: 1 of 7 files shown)


mcdc/input_.py
jpmorgan98, line 1: Take a look at the function header defining domain decomposition for documentation

jpmorgan98 (approved):

mcdc/input_.py
braxtoncuneo: I have updated the documentation to reflect Sam's input.

PR #199, opened 2024-05-24
Remove bilinear source tilting and Davidson's method

spasmann:
Removing these features for maintainability.


ilhamv:
The failing test looks like it's caused by a race condition caused by checking the answer inside `input.py` (which is done by the MPI processes).

What do you think, @clemekay ?

ilhamv:
I'm going to merge this PR and make a follow-up PR to fix the regression test.
Thanks, @spasmann!
(Gotta remember this PR #199, particularly in the case in the future we need the help of a bi-linear tilted iQMC.)

PR #212, opened 2024-07-19
Decomposed mesh tallies when domain decomposition is active

alexandermote:


alexandermote:
Revised mesh tally decomposition after a meeting with @ilhamv:
- Reduced `N_bin` to match decomposed tally values in `main.py`
- Reduced tally sizes in `make_type_mesh_tally` function in `type_.py`

ilhamv:
Thanks, @alexandermote !

The MPI DD tests crashed. But all others passed. Which is a good sign. My recommendation:
(1) Identify the issue/bug in the non-Numba MPI test by manually running it with several ranks (the test uses 4 ranks).
(2) Then move on to the Numba MPI test.
And don't forget back in black (eg, run `black *py` on `MCDC/mcdc`).

alexandermote:
I made several changes with this update:

- Fixed an indexing issue that was causing DD mesh bounds to be set improperly
- Added an `MPI.Gather` call before the HDF5 output is created to re-assemble the mesh tally
- Added an `if not DD` flag to `MPI.Reduce` calls in the tally closeout functions, since each processor is working on a different section of the tally

This should allow at least the non-Numba tests to pass. However, it only works when there is only one processor assigned to each subdomain. I will need to add some version of `MPI.Reduce` among only processors in the same subdomain in order for it to work when multiple processors are assigned to any given subdomain.

PR #213, opened 2024-07-21
Bug fix: particles sourced on cell boundaries

spasmann:
In the updated CSG functions a `particle is lost` error occurred if a particle was sourced exactly on a cell boundary. In standard Monte Carlo, this bug may never surface but samples in iQMC hit these boundaries quite often. 

In short, replacing `=` with `<=` in `region_check()` returns the same cell ID  as was calculated before the updates.


jpmorgan98 (approved):
Simple hot fix!

jpmorgan98:
Thanks @spasmann you rock!!!!!!!!

PR #215, opened 2024-07-25
Bug fix, continuous energy test, and file-based source test

ilhamv:
This PR fixes the CE-physics bug created in the previous commit, adds a test for CE physics (by creating a non-git-tracked dummy continuous energy XS library, and adds a test for file-based source.


ilhamv:
A regression `MCDC/test/regression/slab_ce` test is created for CE physics to avoid similar issues.

ilhamv:
@massimolarsen , I add the file-based source example with this PR as well: `MCDC/test/regression/source_file`

Issue #214, opened 2024-07-24
Visulizer import deprection


jpmorgan98:
I want to change the visualizer to a sub import file so that a user that wants to interface with it will have to do something like `from mcdc import visualizer `

PR #220, opened 2024-08-06
Moved tkinter to try/except in mcdc/visualizer

clemekay:
@jpmorgan98 This isn't quite what you said you wanted to do in #214, but it maybe accomplishes the goals of that? Moving the tkinter module to the existing try/except module in mcdc/vis and adding a sys.exit() to the ImportError exception handling makes it so that any vis-dependency-related errors only print when trying to call mcdc.visualizer(). 

Making this a draft since I'm on a Mac that can't use the dependencies, I can only test the exception handling haha. If someone else could test that the visualizer itself still works, perhaps this small change is a good option?

### DIFF
--- a/mcdc/visualizer.py
+++ b/mcdc/visualizer.py
@@ -1,25 +1,44 @@
+import sys
 from mcdc.print_ import print_warning
 
+
+# Launches visualization window; must be in try/except loop so it doesn't launch upon import
+missing = False
+mods = []
 try:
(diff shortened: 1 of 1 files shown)


jpmorgan98:
I was planing on moving that whole visualizer to a sub-folder such that a user would have to `import mcdc.visualizer`. Kinda like scipy's integration library. That all these errors will show if a user specifically imports them

Issue #206, opened 2024-07-02
GPU compilation fails after PR #198 (next VR)


jpmorgan98:
GPU compilation fails after PR #198. Only resulting in a `Aborted (core dumped)`.  This could be because some of some changing data layout when implementing new tracking method? Also something to note is that the global int flags like `BC` and `FILL_LATTICE:` are never declared and/or moved to the GPU. I am guessing they are supposed to be stored in the global state but that feels like the most significant difference in this PR.

**This must be fixed before next version release**

jpmorgan98:
Fixed as of PR #222

PR #222, opened 2024-08-08
Refactor CSG tracking with Reverse Polish Notation

ilhamv:
The CSG tracking is refactored from a recursion of function calls into an algorithm based on the Reverse Polish Notation (Thanks, @braxtoncuneo, for the suggestions!). This fixes the GPU runtime error (#206) which turns out to be caused by the function recursions.


jpmorgan98:
I am trusting you ran this on the Nvidia system?

ilhamv:
> You ran this on the nivida development system, right?

Yes, I tested it on the mcdc-env you set up in the OSU ssh. But it would be great if you could double-check to be very sure.

jpmorgan98:
Everything passes!

jpmorgan98 (approved):

PR #230, opened 2024-08-15
Adding 3D tally recomposition and 3D Domain Decomposition test

alexandermote:
These changes are mainly improvements to #212, namely expanding the mesh tally decomposition to 3D domain decompositions. A new regression test, `dd_cooper`, has also been added to show and test this capability. Currently, this test passes on mean flux, but fails on standard deviation (see #231).


ilhamv:
Thanks, Alex.

How did you generate the `answer.h5` with which the test passes on the mean value but not the stdev?

alexandermote:
> Thanks, Alex.
> 
> How did you generate the `answer.h5` with which the test passes on the mean value but not the stdev?

I ran the input with domain decomposition turned off. Everything else in the input should be the same.

ilhamv:
I see. So the DD run does not reproduce the stdev of the non-DD run, right?

ilhamv:
> I see. So the DD run does not reproduce the stdev of the non-DD run, right?

... and have you made sure that the same issue only shows up in 3D? Does it also show up in 1D?

ilhamv:
Would it be a good idea to make the `dd_slab_reed` to be tally-decomposed as well? (This would help in debugging as it gives a 1D problem variant)

alexandermote:
dd_slab_reed is already tally-decomposed, as it’s an inherent part of domain decomposition. It passes all tests; I haven’t tested a 1D version of dd_cooper, however.

ilhamv:
I see. We need to make sure if the issue is unique to 3D DD or not: Can you check if `dd_slab_reed` reproduces the non-DD run? I think you can do that by running `dd_slab_reed` manually without DD, saving the output as `answer.h5`, and then running the regression test normally. Thanks, Alex.

alexandermote:
> I see. We need to make sure if the issue is unique to 3D DD or not: Can you check if `dd_slab_reed` reproduces the non-DD run? I think you can do that by running `dd_slab_reed` manually without DD, saving the output as `answer.h5`, and then running the regression test normally. Thanks, Alex.

I ran the test as you described, and got the same issue. Seems like it is tied to DD in general and not 3D DD specifically. I also tested it without the tally recombination introduced in this update, and the issue still occurred, so it was not introduced by these changes.

ilhamv:
The identified issue is not unique to this PR and will be tracked in #231.

PR #243, opened 2024-10-03
Cell Tallies

ethan-lame:
Added cell tally functionality. Can define a cell in the input deck, and call mcdc.tally.cell_tally() to get the tally for the cell. It should properly output to the h5 file. 

Verified it gave the same results as the sum of mesh tally cells using the new sphere_in_cube input deck - a homogeneous pure-fission sphere inside a pure-scattering cube.

(This is my first PR, hopefully I don't break everything!)


ilhamv:
Thanks, @ethan-lame !

Note that the [black formatting](https://black.readthedocs.io/en/stable/) test failed. You can install `black` via `pip` or `conda`, and then do `black .` on the MC/DC top directory to reformat all the files changed. You can make this process automatic at every commit by (1) `pip install pre-commit`, and then (2) run the command `pre-commit install` on the top directory.

Note that the Numba regression tests failed. My recommendation to fix it is to go to `MCDC/test/regression/azurv1` and run the test problem manually on Nubma mode, `python input.py --mode=numba`. It's going to take some time to run as it needs to compile. When it runs, it throws out error messages that would help fix it. Make the necessary fixes until the problem runs without error. Then you go to `MCDC/test/regression/` and run `python run.py --name=azurv1 --mode=numba` to run the regression test only on the selected problem in Numba mode. If it passes, it means most probably the Numba problem has been fixed, and you can make the Git commit and push.

ilhamv:
Thanks, Ethan!

This looks great. Ethan, let's move/copy some of the examples you created as part of the `test/regression` so that the new cell tally capability will automatically be regression tested as well.

ilhamv:
I made minor edits and added little improvement other than the calling of the cell tally scoring in `kernel.py`.
This is good to merge.
(For some reason, my GitHub Review window gets cut out on my browser, and I can't make a Review. I'll just bypass the branch protections and merge it.)

PR #257, opened 2024-11-21
break mesh tally loop on coincidence

ilhamv:
This does not fix a bug. But, this improves some accidental inefficiency. Previously, we always trace the meshes all the way to the problem boundary, even beyond where the particle stops. This is due to the mesh tally scoring function that does not anticipate the new geometry coincidence treatment.


ilhamv:
Closing this because it got mixed up with other things. Separate PRs will be made.

PR #266, opened 2024-12-19
Compressed Sensing Implementation

ethan-lame:
Added compressed sensing methods to MC/DC - this includes new functions to find/perform the reconstruction and a new `mcdc.tally.cs_tally` to be used in the input file, similar to `mcdc.tally.mesh_tally`. MC/DC automatically generates the reconstruction and writes it to the output file - the `examples/fixed_source/sphere_in_cube/process.py` file can also use just the compressed sensing data to perform a more bespoke reconstruction (e.g. for testing different solvers). As of writing this, I've only tested it on the `sphere_in_cube` example.

It does not yet work in numba mode.


ilhamv:
Thanks @ethan-lame!

I have verified the Numba issue is in `calculate_distance_in_coarse_bin`. (I commented out the function call and Numba works)
I think the issue is caused by the functions defined within that njit-ed function. Is it possible to define the functions independently (and also njit-ed independently)?

ethan-lame:
I think I found the problem with the Numba issue - I was trying to redefine cs_bin_size in the score_cs_tally function when using the last bin. I changed that and it ran, but I'll still have to do some testing to make sure it's accurate.

ilhamv:
Looking good, Ethan!
I'm currently fixing the issue of the failing MPI tests. Note that the issue is not specific to this PR.

ilhamv (approved):

PR #273, opened 2025-01-20
Fix bug in time census

ilhamv:
There is a bug in the time census. A line was accidentally removed from the code. The bug has not been detected, as there is no regression test for the time census. Thus, this fix comes with the regression test.


jpmorgan98 (approved):

jpmorgan98:
Actually can you change [this line](https://github.com/CEMeNT-PSAAP/MCDC/blob/dff481976e27a4efb297d17ab3bfc07173188d85/pyproject.toml#L51) to requires numba>0.60.0

ilhamv:
Good point. After the test passes, I'll add that and then merge it in.

PR #274, opened 2025-01-20
Debug workflow

### LINKED ISSUE #263: Remove Numba patching
I just manually installed MC/DC on a new clean conda environment with `pip install .`. The installation works, even in Numba mode. Does that mean we don't need to patch Numba anymore? Can anyone verify that this is the case? If it is, we can just remove the Numba patching scripts.

Note: I installed it on Dane. The listed Numba version is 0.60.0

ilhamv:
The Numba patch is no longer necessary per recent Numba updates. In the latest update, Numba 0.61.0, the patch creates errors. This PR removes Numba patching.

Fix #263


ilhamv:
Note that rerunning a previously successful regression test on the current `dev` branch causes the test to fail.

PR #275, opened 2025-01-21
Upgrade time census

ilhamv:
This PR adds several upgrades related to the time census capability.

# Particle Future Bank
The future bank stores particles beyond the current census time. The future bank is typically filled with fixed-source particles and delayed neutrons emitted in the future beyond the next census time. These future particles are separated from the censused particles so that they are not prematurely subjected to population control. Additionally, similar to the other banks---active bank, census bank, and source bank---future bank buffer may need to be allocated for certain problems.

# Census-based tallying
This overrides the tally time grid, which is based on the census time grid and a user-defined frequency. For example: `mcdc.time_census([1.0, 2.0, 3.0], tally_frequency=2)` generates tally time grid of `[0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0]`. Only tally bins that are relevant to the current census are stored, so that this avoids storing the entire time-dependent tally bins.


ilhamv:
@shac170 , The PR is ready for you to look at. Let me know what you think.

ilhamv:
Note: This capability is needed to run the time-dependent SMR challenge problem.

shac170:
This seems good to me, the census tallying options appear to work and will be very useful. It could also be nice to have the ability to combine census output files into a single file after the simulation is complete, though this could be done through post-processing.

shac170 (approved):
This seems good to me, the census tallying options appear to work and will be very useful. It could also be nice to have the ability to combine census output files into a single file after the simulation is complete, though this could be done through post-processing.

ilhamv:
@shac170 - Yes, it would be great if we have something like `mcdc.combine_census_tally(time=..., output_name=..., ...)` as a post-processing tool. For now, the regression test `azurv1/tally_census` provides a post-processing example for merging the census tally outputs into a single output file.

PR #282, opened 2025-01-27
Feature/northroj/time cell tallies

northroj:
This PR adds time and energy binning to cell tallies Ethan added recently.

The structure was primarily copied from the mesh tallies. Domain decomposition support is currently not included. It is possible/probable that cell tallies crossing domain decomposition borders will cause errors, but it is unlikely that cell tallies would be used in the types of large problems where domain decomposition is used.

This has been tested on an approximate pulsed sphere problem currently being added to the challenge problem suite in conjunction with time census.


northroj:
I also updated the coincidence tolerance in this PR. I have not yet run into a geometry which causes this tolerance to fail, but I will keep looking.

northroj:
I also updated the coincidence tolerance in this PR. I have not yet run into a geometry which causes this tolerance to fail, but I will keep looking.

ilhamv:
Thanks, Jordan!

There are possible cases in which one would need a cell tally and domain decomposition at the same time. However, the cell tallies are less likely to be the kind of tally that needs decomposition. So, for now,  this should be OK in most cases.

I plan to review the PR soon tomorrow.

ilhamv:
It was weird that the MPI-Numba test passed, but the MPI-Python didn't. Re-running the latter now.

ilhamv:
@northroj - can you add a regression test for the time-meshed cell tally feature?

ilhamv:
@northroj - I really like your clear and complete input parameter descriptions in `def cell_tally()`. Can you update the other tally parameter descriptions? Thanks.

ilhamv (approved):

northroj:
> @northroj - I really like your clear and complete input parameter descriptions in `def cell_tally()`. Can you update the other tally parameter descriptions? Thanks.

I am not sure what this means. The docstring in def cell_tally() was copied from the other tally docstrings, and I don't know what other tally parameter descriptions to update.

northroj:
I just realized I cant use continuous energy for the regression test, I will get a mg one in.

northroj:
For @ilhamv: For a test of the cell tallies, I just added a cell tally to the inf_shem361 problem along with census to test the time and energy binning along with time census. I had to comment out the mesh tally because it produced a NAN for the stdev in the final energy group bin when census was enabled (it does not show up when census is commented out). I will add this to an issue.

ilhamv:
Oops. Looks like I messed up the commit history... Let me try fixing it.

PR #284, opened 2025-02-04
Fixing Docs Deployment and adding GPU computation notes

jpmorgan98:
Docs are currently failing and I had to add `cvxpy`
I also include a description of MC/DC compiling to GPUs


clemekay (commented):
Successfully built on readthedocs, but Numba/MPI test fails

jpmorgan98:
Looks like it is passing now--weird

clemekay (approved):

sellenth:
JP, looks like some stellar work here.

PR #295, opened 2025-02-18
Automatic Lagged Weight Windows

shac170:
This pull request includes the ability to use automatic weight windows based on the previous timestep (lagged) solution. This formulation creates its own weight window tally based on the provided mesh and uses the census based tally outputs to assign weight windows. This must be used with time census. An example file is included.


ilhamv:
Thanks Caleb @shac170 !

Per the discussion in the dev meeting:

- [ ] Add regression tests
- [ ] Wait for @clemekay's review (as the PR serves as good data to consider in the code modularization plan)
- [ ] Wait for the current GPU bug in `dev`, #300 , to be fixed (I'm working on that), so that then we can assess the PR's GPU compatibility

ilhamv:
@shac170 - The new regression test seems to only work in Python mode. Can you please check that?

shac170:
> @shac170 - The new regression test seems to only work in Python mode. Can you please check that?

@ilhamv this is very strange, it seems to work in Numba mode for me on my computer...

ilhamv:
We will work on merging this PR after the GPU bug fix, #310, is merged in (very soon).

ilhamv:
I have confirmed that the GPU regression test passed.

PR #301, opened 2025-02-24
adding recombine_tallies and bug fixes

shac170:
Added a function to recombine the census based tally outputs. Also fixed two bugs with this type of tallies.


ilhamv:
The first bug was when there were several mesh tallies, they were overwriting each other so only the last one would appear in the output. This is fixed by appending instead of writing after the first tally. The second bug appears when using the census tallies without population control. The runtime from batch to batch significantly increases. This is because `mpi_work_size` was not reset at the beginning of each batch. The bug can be reproduced by turning off population control in one of the relevant regression tests.

ilhamv:
Thanks, Caleb @shac170.

The fixes seem to change the result and fail the regression test of `azurv1_census_tally`. Can you please verify that, @shac170?

Also, let's use the new tally recombination function in the regression test `azurv1_census_tally`.

ilhamv (approved):

PR #316, opened 2025-03-23
Update regression test

ilhamv:
Making all the fixed source problems run in batch statistics to get stdev reproducibility on GPU.


ilhamv:
Hi @shac170 - I just noticed that we have the weight window input setting in `azurv1_ww` commented out. I just want to make sure if that is on purpose.

clemekay:
@alexandermote Here's the PR where the non-reproducibility issue is popping up, regression doesn't pass for mean or sdev. 

I know in `loop.py` (line 72) at the start of the batch loop, the `distribute_work` function doesn't get called if domain_decomp is turned on, but I'm not familiar enough with the domain decomp functionality to know if what happens in `distribute_work` would be happening elsewhere for domain decomp:

```
    # Loop over batches
    for idx_batch in range(mcdc["setting"]["N_batch"]):
        if not mcdc["technique"]["domain_decomposition"]:
            kernel.distribute_work(N=mcdc["setting"]["N_particle"], mcdc=mcdc)
        mcdc["idx_batch"] = idx_batch
        seed_batch = kernel.split_seed(idx_batch, mcdc["setting"]["rng_seed"])
```

alexandermote:
After looking things over and running a couple of tests with Kayla, we're pretty sure we've narrowed the reproducibility issue down to the tally code that runs at the end of multi-batch runs. I added some DD functionality to those functions a while back, and I don't think my additions are properly set up for multi-batch runs. Currently working on a fix.

alexandermote:
I think the answer key for `slab_reed_dd` needs to be updated; the differences in the regression test now look like the difference between a 1-batch and a 2-batch run. I'm able to get the test to pass using non-DD 2-batch results as the answer key on my local install.

ilhamv:
That sounds good, @alexandermote. Please go ahead and update the answer key for `slab_reed_dd`. Thanks!

alexandermote:
@ilhamv I don't have permission to upload files to your repository; if you give me push access, I can upload the new answer key. Thanks for your help!

ilhamv:
@alexandermote - I think I added the permission for you; please go ahead and try it.

This PR fixes #318

clemekay:
@ilhamv , would you mind running these on GPU again? I'm just curious to know whether it'll pass before merging. All CPU tests pass!

ilhamv:
All regression tests pass on GPU, except the `source_file` problem, which is expected as the tested feature does not support batch run yet. I'll open up an Issue for that. But the PR is good to merge.

ilhamv:
When I said that the tests passed, that includes the standard deviation reproducibility, which was included by the most recent commit.

PR #319, opened 2025-04-11
Data refactoring

ilhamv:
Refactoring some critical data (tally, banks, etc.) to their dedicated numpy arrays.


ilhamv:
# Regarding the need for Tally refactoring

The previous tally data layout is set up as the following:
```
     # Set tally data
     if not input_deck.technique["uq"]:
         tally = np.zeros((3, tally_size), dtype=type_.float64)
     else:
         tally = np.zeros((5, tally_size), dtype=type_.float64)
 
     # =========================================================================
     # Establish Data Type from Tally Info and Construct Tallies
     # =========================================================================
 
     type_.make_type_tally(input_deck, tally_size)
     data_arr = np.zeros(1, dtype=type_.tally)
```

Instead of being represented as a standard 1D Numpy array (the one defined per `tally =`, which is no longer used at all), tally data is now represented as a single-element structured array (the one defined as `data_arr`). If I remember correctly, we went on this path because of some mutability challenges in GPU mode. Is that correct @braxtoncuneo?
Nevertheless, this new path calls back our past limitation that shows up when the tally is super large (`tally_size` is very large), raising the following error:
```
  File "/g/g92/variansyah1/MCDC/mcdc/type_.py", line 164, in into_dtype
    result = np.dtype(align(field_list), align=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
```

Do you think it will work OK if we just define the tally data array as its own standard Numpy 1D array, like what is implemented in this PR, @braxtoncuneo? (with the drawback of the need to explicitly pass it as a function argument)

Note that the GPU mode does not work. I tried to fix it, but it got deeper into `harmonize`...

PR #328, opened 2025-04-24
Tally refactoring

ilhamv:
Refactoring tally to their dedicated numpy arrays.

The previous tally data layout is set up as the following:
```
     # Set tally data
     if not input_deck.technique["uq"]:
         tally = np.zeros((3, tally_size), dtype=type_.float64)
     else:
         tally = np.zeros((5, tally_size), dtype=type_.float64)
 
     # =========================================================================
     # Establish Data Type from Tally Info and Construct Tallies
     # =========================================================================
 
     type_.make_type_tally(input_deck, tally_size)
     data_arr = np.zeros(1, dtype=type_.tally)
```

Instead of being represented as a standard 1D Numpy array (the one defined per `tally =`, which is no longer used at all), tally data is now represented as a single-element structured array (the one defined as `data_arr`). If I remember correctly, we went on this path because of some mutability challenges in GPU mode.
Nevertheless, this new path calls back our past limitation that shows up when the tally is super large (`tally_size` is very large), raising the following error:
```
  File "/g/g92/variansyah1/MCDC/mcdc/type_.py", line 164, in into_dtype
    result = np.dtype(align(field_list), align=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
```

Per discussion with @braxtoncuneo, this change is going to work. However, some work needs to be done before the GPU mode can run.

Note that this refactoring is required to run the larger problems, like high-fidelity C5G7 and SMR 4-phase problems.

Question: Should we merge it now and later make a separate PR for the GPU fix, or should we wait until the GPU fix is in?

### DIFF
--- a/mcdc/kernel.py
+++ b/mcdc/kernel.py
@@ -1850,9 +1850,8 @@ def mesh_get_energy_index(P_arr, mesh, mode_MG):
 
 
 @njit
-def score_mesh_tally(P_arr, distance, tally, data, mcdc):
+def score_mesh_tally(P_arr, distance, tally, data_tally, mcdc):
     P = P_arr[0]
-    tally_bin = data[TALLY]
     material = mcdc["materials"][P["material_ID"]]
     mesh = tally["filter"]
     stride = tally["stride"]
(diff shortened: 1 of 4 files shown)


clemekay:
Per @braxtoncuneo, we will merge this in now and do a separate PR later for the GPU capability

PR #329, opened 2025-04-24
Delta tracking

### LINKED ISSUE #335: Hybrid Delta Tracking
From the 05/07/2025 document "MC/DC Plans Through the End of CEMeNT" :

> Merging my PR into dev and adding documentation on how to use exactly. I also want to streamline some functions, make some logic clearer and implement unit and regression test as well as adding an example. More stretch goals would be to enable delta tracking with k-eigenvalue calculations and show the collision estimator some love. I can expand on this if needed
> I plan to start this after my defense sometime in the summer

jpmorgan98:
Adding a bunch of new tracking features to explore various kinds of hybrid delta tracking schemes. Most of the meat and potatoes is in main.prepare where the majorant is computed and `kernel.move_to_event` where delta tracking actually happens.

Note that I am adding a `bool` to each particle to keep track of weather or not it is undergoing delta tracking. This feels like kinda a big deal

# New Features

- A collision estimator
- Functions to compute both the macroscopic and microscopic majorants
- Ability to plot the majorant
- Rejection sampling
- Visualizer improvements with the ability to save time outputs as `.gifs` and print legends
- Hybrid methods including an energy cut off and material exculsion
- Adding documentation functionality to use

# Some Todos
- Make delta tracking on/off logic in `kernel.move_to_even`t a little more streamlined
- Make a separate `compute_majorant` functionality to plot the majorant without having to run the simulation
- Giving materials input-able names so they can be easier to plot and identify 
- Allowing multiple material regions to be excluded (right now it's only 1)

# Verification
Passing verification sueit, also producing the correct results with C5G7, C5CE, Kobyashi, and Dragon all of which see speedups when delta tracking is enabled. Or hybrid delta tracking.

# New dependency
I am adding a new dependency so that the visualizer can make the movie. Its based on matplotlib and is pretty light weight.

# Limitations
This delta tracking does not support fission tallies so also is not supported in k-eigenvalue computations


ilhamv:
Thanks, @jpmorgan98. Note that in k-eigenvalue, we can use the collision estimator to get the estimate of the next iterate of k.

Issue #346, opened 2025-06-30
mpi4py CI error


jpmorgan98:
The `mpi4py` github runner seems to be having issues again. This is likely related to issue #267 

The all test which attempt to install and run MC/DC fail in the mpi4py setup process yielding this error:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/mpi4py/bench.py", line 792, in <module>
    main()
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/mpi4py/bench.py", line 779, in main
    from . import MPI
ImportError: libmpi.so.12: cannot open shared object file: No such file or directory
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/mpi4py/bench.py", line 792, in <module>
    main()
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/mpi4py/bench.py", line 779, in main
    from . import MPI
ImportError: libmpi.so.12: cannot open shared object file: No such file or directory
Error: Process completed with exit code 1.
```

jpmorgan98:
Ok setting the CI version to `ubuntu-latest` seems to solve the problem. Also it seems to suggest that the issue in #267 is no longer an issue. Adding this to a PR for review.

ilhamv:
The issue is no longer evident in the runner executions of the current code version.

PR #347, opened 2025-06-30
Changing CI OS version from ubuntu-22.04 to ubuntu-latest

jpmorgan98:
To address issue #346


ilhamv:
It is possible that this has been solved in the refactored code. Closing it, for now.

PR #350, opened 2025-09-08
Refactoring kernel.py's tally_accumulate function for potential 3x speedup on pure Python

jburz2001:
Hello. I was curious about which parts of the _slab_absorbium_ example contributed the most to the wall clock time of calling `mcdc.run()` and I might have made the pure Python version much faster. Old code is currently temporarily commented out in the commits being PR'd here in case someone wants to verify each. I'm hoping someone could verify my results 🙂

First, I created a copy of the _slab_absorbium_ example in a new directory called _slab_absorbium_profiling_, which uses the cProfile profiler to identify timing bottlenecks. Then I re-wrote the input.py file for this example to examine the runtime of `mcdc.run()` after having first "warmed up" Numba JIT compilation. When I run it for a 200 particle "warm-up simulation" and a 1e3 particle "warmed up simulation" I get the following output (the warm-up isn't necessary here though in pure Python mode) :
mcdc_outputs_original

I also used cProfile to output the top 25 functions that contributed the most to wall clock runtime:
cProfile_output_original

It's clear from cProfile that the function `tally_accumulate(data, mcdc)` in `kernel.py` contributes significantly to runtime. Notably, the original function leverages a for loop for "Numpy vectorizable" operations. Thus, one one can rewrite the code to circumvent explicit "for looping", yielding faster runtime on pure Python. That change is what this PR commits to the codebase.

I reran the example with the new function and get the following output:
mcdc_outputs_updated

Additionally, when viewing the new output from cProfile, we see that `tally_accumulate(data, mcdc)` is no longer a significant contributor to runtime:
cProfile_output_updated

Most importantly though, all of the regression tests from the _/MCDC/test/regression_ directory still pass, but it would be nice if someone could verify/validate this updated code too:
test_results_updated

Also, the output from process.py looks the same:
process_results_updated

I then tested the runtime with 1e4 particles too and observe ~3x speedup during pure Python mode and a slight speedup in Numba mode (1e5 isn't plotted here but I observed the same ~3x speedup for pure Python). The text in black is the mean wall clock runtime in seconds; standard deviation bars are plotted yet are too small to see on this log-scale:
Untitled

### DIFF
--- a/examples/fixed_source/slab_absorbium_profiling/input.py
+++ b/examples/fixed_source/slab_absorbium_profiling/input.py
@@ -1,0 +1,60 @@
+import numpy as np
+import time
+import cProfile, pstats
+
+import mcdc
+
+
+def build_model(N_particle):
+    # materials
+    m1 = mcdc.material(capture=np.array([1.0]))
+    m2 = mcdc.material(capture=np.array([1.5]))
+    m3 = mcdc.material(capture=np.array([2.0]))
+
+    # surfaces
+    s1 = mcdc.surface("plane-z", z=0.0, bc="vacuum")
+    s2 = mcdc.surface("plane-z", z=2.0)
+    s3 = mcdc.surface("plane-z", z=4.0)
+    s4 = mcdc.surface("plane-z", z=6.0, bc="vacuum")
+
+    # cells
+    mcdc.cell(+s1 & -s2, m2)
+    mcdc.cell(+s2 & -s3, m3)
+    mcdc.cell(+s3 & -s4, m1)
+
+    # source
+    mcdc.source(z=[0.0, 6.0], isotropic=True)
+
+    # tally
+    mcdc.tally.mesh_tally(
+        scores=["flux"],
+        z=np.linspace(0.0, 6.0, 61),
+        mu=np.linspace(-1.0, 1.0, 32 + 1),
+    )
+
+    # settings
+    mcdc.setting(N_particle=N_particle)
+
+
+def run_once():
+    t0 = time.perf_counter()
+    mcdc.run()
+    return time.perf_counter() - t0
+
+
+if __name__ == "__main__":
+    # warm-up JIT compiler
+    build_model(200)
+    _ = run_once()
+
+    # use warmed-up JIT compiler
+    build_model(1e3)
+    pr = cProfile.Profile()
+    pr.enable()
+    wall = run_once()
+    pr.disable()
+    print(f"Wall time (profiled run): {wall:.3f} s")
+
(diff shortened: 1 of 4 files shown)


ilhamv:
Hi @jburz2001 - Thanks for bringing this up.

I see. So, numpy vectorized operations are much faster than the manual for-looping. And I see that is the case because Numpy uses a compiled implementation and potential parallelism under the hood. I believe the speedup would be much more extreme if we had a bigger tally (very large `N_bin`)!

However, I'm wondering if Numpy vectorized operations are supported by Numba's `cuda.jit`... Nevertheless, given the operation's potential significance on runtime, perhaps a GPU-adaptive `tally_accumulation` function would be a good thing to have.  Thoughts on this @braxtoncuneo @jpmorgan98?
(And @jburz2001 - this is one of the things that is unique in MC/DC development, the core capabilities need to accommodate all the way from Python mode to the GPU mode.)

Another thing to mention is: If you look at the profiler results, you may find that the `tally_accumulate` function is called as many times as the number of particles. So, it's a lot! However, in practice, when we run fixed-source simulations, to get the statistics, one would typically run the simulation in batches. Such that, if in the setting you add `N_batch=2`, for example, you may find that `tally_accumulate` is no longer in the top parts of the hot spot list.

ilhamv:
Justin @jburz2001, these ideas to optimize MC/DC are very useful. I'd like to point you to this MC/DC branch: https://github.com/CEMeNT-PSAAP/MCDC/tree/data_refactor. That is where the major refactoring of MC/DC is happening (the plan is to eventually merge it into `main` and `dev`; so, any thoughts/ideas on that branch are appreciated, too. 😄

jburz2001:
Yup, Numpy vectorized operations tend to be much faster than manual for-looping. This is partly because Numpy is optimized for doing such operations but also because pure Python isn't particularly performant as a dynamically-typed, interpreted language. For instance, during _each_ iteration of the loop, Python's standard interpreter (called CPython) has to check variable types to ensure valid arithmetic, adding significant overhead to something that looks to us like obviously-valid arithmetic!

I suspect that the `@vectorize` decorator with `target=cuda` would be a good place to look for creating jitted, CUDA-compatible, vectorized operations that maintain Numpy syntax. Something like
```
@vectorize(['float32(float32, float32)'], target='cuda')
def add_ufunc(a, b):
    return a + b
```
which I took from [this site](https://carpentries-incubator.github.io/gpu-speedups/01_CuPy_and_Numba_on_the_GPU/index.html).

I'll take a look at the data_refactor branch 🙂.

Issue #330, opened 2025-04-30
Add support for py 3.13


MicahGale:
I sadly got the error:

``` bash
Processing /home/mgale/dev/MCDC
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
INFO: pip is looking at multiple versions of mcdc to determine which version is compatible with other requirements. This could take a while.
ERROR: Package 'mcdc' requires a different Python: 3.13.3 not in '<3.12,>=3.9.1'
```
Hopefully adding 3.13 is just adding testing for this version and updating `pyproject.toml`. I can take a crack at this... if `numba` supports this.

jpmorgan98:
They just keep coming and they don't stop coming! We put in the hard limit as Numba was slow to adopt Python/3.12. I will run our full verification testing suite on Python/3.13 (with GPU compilation) later this week and see if everything is working on 3.13. Thanks for bringing this up @MicahGale

MicahGale:
Thanks for looking into this. I do like being a bit too much on the bleeding edge. (I PR'd Py 3.13 support for OpenMC on Py 3.13 release day I think). New python releases come out every October. It looks like the first version of [numba](https://pypi.org/project/numba/0.59.0/#history) to support py 3.13 was `0.61.0` released this Jan 20, and added py 3.12 support in `0.59.0` on Jan 31, 2024. So it seems like February would be a good time to evaluate adding support for new python versions.

PR #353, opened 2025-12-10
Patches for python 3.12+

jpmorgan98:
I want to run this on my mac so I can do some work. I had to make some changes to the polymorphic class but the regression tests seem to be fine. Hopefully this doesn't screw anything else up for folks. ChatGPT helped!

### DIFF
--- a/.github/workflows/black_lint.yml
+++ b/.github/workflows/black_lint.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: "3.13"
           cache: pip
 
       - name: Install Black
--- a/.github/workflows/black_lint.yml
+++ b/.github/workflows/black_lint.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
 
       - name: Install Black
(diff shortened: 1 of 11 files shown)


ilhamv:
Thanks for getting MC/DC compatible with Python 3.13, @jpmorgan98 !

Some tests failed at the MC/DC installation step. I thought that it was because of the Python version used in the workflow, so I bumped it up to Python 3.13; but the error was still there.

Turns out it is due to the `Windows` classifier. After I commented it out, it works.

1. Do we need `Windows` in the classifier? If we do, I think some work may need to be done. (particularly because I got the same error locally on Macbook)
2. The workflow currently tests for Python 3.13. Should we also workflow-test all the way from Python 3.9 as described in the `pyproject.toml` requirement? Or, we can just make the requirement tighter: `requires-python = 3.13`

jpmorgan98:
Removed windows. I say no to locking in on a specific version of python. Many of our HPC envs are Python 3.10. I don't know where that leaves our continuous integration. I say we test 3.10.0 3.11.0 3.12.0 and 3.13.0 and current upon version release and latest on puch

jpmorgan98:
All tests should pass!

jpmorgan98:
This PR solves Issue #330

ilhamv:
Great!

I set the workflow to test on 3.10 and 3.13 through `matrix`, and I revised the Python version requirement to be >=3.10.
I found that it does not work for 3.9. You mentioned that many HPCs use 3.10. So, I guess we are still good?

Aldo, to reduce workflow description redundancy, I merged the different workflow descriptions of the four modes (Python/Numba, Serial/MPI) into one through `matrix`.

ilhamv (approved):

ilhamv:
@jpmorgan98 - If you think the workflow is good to go now, please go ahead and merge. Thanks!

PR #357, opened 2025-12-18
Moving source

ilhamv:
The moving source uses a very similar approach to the moving surface, and that is translating the particle position based on the static object (source/surface).

An example and a regression test are added to demonstrate and test the new capability.

A snippet example to move a source (from `MCDC/examples/moving_source/input.py`:
```
src = mcdc.Source(
    x=[-4.0, -3.0],
    y=[-0.5, 0.5],
    z=[-0.5, 0.5],
    isotropic=True,
    energy_group=0,
    time=[0.0, 10.0],
)
src.move([[1.0, 0.0, 0.0], [-0.5, 2.0, 0.0], [0.0, -3.0, 0.0]], [7.0, 2.0, 1.0])
```

This gives the following:
moving_source


mcdc/object_/source.py
```diff
@@ -203,3 +220,36 @@ def __repr__(self):
             text += f"  - Time: {self.time_range} s\n"
 
         return text
+
+    # ==================================================================================
+    # Source moving
+    # ==================================================================================
+
+    def move(self, velocities, durations):
```
jpmorgan98, line 228: Should this be capitalized to follow other new input cards like `Materials` and `Surfaces`?

jpmorgan98 (approved):
Ahhh this is dope!!!!!

I have a question about files in `mcdc_get` and `mcdc_set`. Each file is headed with `# The following is automatically generated by code_factory.py` do we want to be committing these files?

mcdc/object_/source.py
ilhamv: Yes, we capitalize for the classes, like `Material` and `Surface`. The `move` here is a method of the class. So, we want to keep it lowercase. Is that what you're asking here, @jpmorgan98 ?

ilhamv: About `mcdc_get` and `mcdc_set`: They are generated by the code factory based on all the items in `mcdc/object_`. So, it is not problem-dependent and changes only if an object description changes or new objects are added.

There is a bit of a hiccup that when the code factory generates them, there will be a slight difference due to an end-of-line white space thingy, and Git will notice. But once it is Black-linted, it will be consistent again. If we can fix that hiccup, it would be great.

PR #359, opened 2025-12-24
Continious energy generator bugfixing

jpmorgan98:
Fixes some issues with the continuous energy generation tooling and adds more robust error handling when things go awry. Adds support for all ACE file interpolation rules and energy sampling techniques including `ACEtk.continuous.MultiDistributionData` as a results runs pretty slow. 

This may need to be accelerated in the future (multiprocessing)


ilhamv:
Thanks, @jpmorgan98 .

I see that the PR passed all checks at one point, and then it failed after the next push. Is it still a work in progress?
Also, does this PR also contain new transport features (just to be sure, as the branch is named `gear_mc`) beyond the continuous energy generator tools?

PR #360, opened 2026-01-13
Visualizer Numba Optimization

gunnarrl:
The visualizer now works with numba, the speedup is 20x-100x+, depending on resolution and number of plots generated. The generated plots are identical to before, and the API is unchanged.

It uses the same algorithm, using `locate_particle()` to get the material_ids, but the inner loop was extracted to work with numba. I had to move it to a separate module because of issues with circular imports from `mcdc.transport.geometry` in `main`. Now there is a wrapper function in `main` that calls the actual implementation in `visualize`. 

Also sped up the plot creation by using color mapping and added caching of `preparation()`.

Speedup (numba already compiled):
| Resolution | Before Time (s) | After Time (s) | Speedup |
|------------|-----------------|----------------|---------|
| 50x50      | 0.874           | 0.048          | 18.2x   |
| 100x100    | 2.131           | 0.052          | 41.0x   |
| 200x200    | 4.601           | 0.073          | 63.0x   |
| 500x500    | 16.271          | 0.147          | 110.7x  |
| 1000x1000  | 67.135          | 0.360          | 186.5x  |


ilhamv:
Hi @gunnarrl !

I apologize for just responding to this PR. I appreciate your patience.

Thanks for upgrading the visualizer. It is blazing fast now - awesome!

I made a minor edit to streamline the imports. I agree that having a separate `visualize.py` file improves the code base organization. Again, thanks, Gunnar!

PR #366, opened 2026-02-07
fix table selection in sampling

ilhamv:
As discovered by @melekderman, and also implemented in #355.
Need to integrate it sooner, as it is also essential for neutron.


ilhamv:
@melekderman - Can you please review if it's all good? Thanks!

melekderman (approved):
@ilhamv Looks good to me, thank you.

ilhamv:
Thanks, @melekderman.

Will merge this after #367.

PR #367, opened 2026-02-07
move in Numba-compiled visualization subfunction into the main function

ilhamv:


ilhamv:
I didn't see the potential import issue that was addressed by @gunnarrl in #360. The root of the issue comes from the non-created `transport/literals.py`, which is engaged by my added commits in that same PR. The issue didn't show up when it was run locally because of the pre-existing `literals.py`; that's why I rushed-merged the PR (sorry!).

In this PR, I address the issue differently by moving the Numba-compiled subfunction into the main function. This would redefine the subfunction whenever `visualize` is called, which I think is necessary, as the key transport function `locata_particle` needs to be recompiled anyway whenever the model, and thus `literals.py`, is changed.

What do you think, @gunnarrl?
Also, that being said, would the cached simulation be invalid if the model changes? That means, should we just always run `preparation` just to be safe?

gunnarrl:
@ilhamv sounds good! 

For the cache, I assumed that the visualization calls would be after the entire simulation is set up, and I was just trying to get it as fast as possible. If the simulation changes after the first `visualize` call, then the cache would be outdated. Probably best to just remove it and call `preparation` every time.

PR #388, opened 2026-03-24
Add neutron inelastic scattering law unit tests

massimolarsen:
Adds basic unit tests for the neutron inelastic scattering laws from the MCNP 6.3.1 manual (https://mcnp.lanl.gov/manual.html) in transport/distribution.py. 

Each test uses a set of test data with known distributions and deterministic rng sequences to compute expected analytic values of E and mu. The expected values are then compared with the sampled values from MCDC.

All the variable names and equations are pulled from the MCNP manual to hopefully keep it readable.

Currently includes: 

- Law 3: level scattering
- Law 4: tabulated distribution
- Law 4: multi-table distribution
- Law 7: simple Maxwell fission spectrum
- Law 9: evaporation spectrum
- Law 44: Kalbach-87 correlated energy-angle scattering
- Law 61: tabulated correlated energy-angle scattering
- Law 66ish: reduced N-body sampling (just checks the current implementation)

Verify with:

`pytest test/unit/transport/distributions`

### DIFF
--- a/test/unit/run.py
+++ b/test/unit/run.py
@@ -5,12 +5,20 @@ from pathlib import Path
 root = Path(".")
 paths = root.rglob("*.py")
 
-EXCLUDE = {"run.py", "__pycache__", ".git"}
+EXCLUDE = {"run.py", "__pycache__", ".git", "conftest.py", "__init__.py"}
+EXCLUDE_PREFIX = {"make_test_"}
 
 start = time.perf_counter()
 for path in paths:
+    # Skip exact matches
     if any(part in EXCLUDE for part in path.parts):
         continue
+
+    # Skip if any part starts with any prefix in EXCLUDE_PREFIX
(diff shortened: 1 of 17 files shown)


ilhamv (approved):
This is great, @massimolarsen - thanks!

I made minor edits to pass the Black-editing. It is possible that the Black version/parameters that you used are different from what is used in the MC/DC GitHub workflow. Note that the default line-break limit that we use is 88 characters.

Additionally, I include the new tests in the runner script `MCDC/test/unit/run.py`, so that it will be automatically run in the GitHub workflow.

We may also need to consider different cases in the tests, including edge cases, for each of the tested distribution evaluations. Those can be follow-up PRs.

I'll merge this PR now, as it sets up the foundational pieces for the distribution test. Therefore, by following the set reference, everyone can collaborate, adding more cases (and new distributions).

PR #396, opened 2026-04-13
Redesign and track numba_types; exclude black-formatting for scripts generated by code_factory

ilhamv:
`code_factory` generates `mcdc_get/`, `mcdc_set/`, and `numba_types.py`. The last one is currently not git-tracked because it is problem-dependent. Unfortunately, this problem-dependent behavior and literal import hinder running multiple instances of MC/DC simulation simultaneously.

This PR redesigns `numba_types.py` so that it is no longer problem-dependent.
Additionally, we exclude black-formatting check for the three files generated by `code_factory`.


mcdc/code_factory/numba_objects_generator.py
```diff
@@ -1206,3 +1242,20 @@ def singular_to_plural(word: str) -> str:
         parts[-1] = w + "s"
 
     return "_".join(parts)
+
+
+def decode_structure_item(item, prefix=""):
```
nglaser3, line 1247: Can you add a docstring or comment to explain what is contained in `item`?

nglaser3 (approved):
Thank you @ilhamv! The changes to `numba_types` looks good to me, and thank you for fixing the black changes to `mcdc_get` and `mcdc_set`! I have one request to add a docstring or comment to the `decode_structure_item` method in `numba_objects_generator.py` just to clarify what the indices into the `item` are, but otherwise approve. If you do not think the docstring / comment is needed since the method is in a part of the code that most people will never touch, that is ok with me and I will merge as is :)

mcdc/code_factory/numba_objects_generator.py
nglaser3: Actually I take this back I do not think it is needed

ilhamv: Thanks, Nathan. I really like informative docstrings and hinted functions. Codes are read more than written, and we can really use more of them in MC/DC! Thanks for bringing this up; I agree that the function could benefit from a docstring.

Issue #345, opened 2025-05-08
MI300A unified memory management


clemekay:
From the 05/07/2025 document "MC/DC Plans Through the End of CEMeNT" : 

> MI300A unified memory space allows for a single copy of GPU data to be shared between the host (CPU) and device (GPU). Enabling this functionality would result in dramatic memory savings (upto 2X) and would enable us to run our challenge problems on fewer nodes
>
> The very same functionality that would support the conveyance of larger data structures to the GPU would additionally facilitate the use of unified memory on MI300A.

PR #401, opened 2026-04-16
CUDA+ROCM Execution Functionality

braxtoncuneo:
## Summary of changes
This PR incorporates functionality that enables execution on both ROCM and CUDA platforms.

## Types of changes

Instead of evaluating the tally shape as a literal in `eval`, and assigning the result to a local variable, the entire assignment is ran through `exec` and the local variable in the resulting namespace is extracted to create the literal value.

While the original form worked for AMD, the other form is apparently required for CUDA to recognize the value as a literal. Perhaps the lack of assignee in the `eval` means that numba cuda doesn't recognize it as something that needs to be literalized.

## Associated Issues and PRs
With this, and the recent merge of `amd-nvidia-interop` into `hamonize/main`, we can probably mark #345 as complete.


braxtoncuneo:
This PR is being closed. Since the change itself is so simple, it makes sense to merge into `dev` instead.

ilhamv (approved):

PR #426, opened 2026-05-15
CI Improvement

melekderman:
## Summary of changes

Cleaned up both workflows in this PR.

## Types of changes

***Regression tests workflow:***

- **Triggers:** Refactored `push`/`pull_request` to `[main, dev]` to stop double-running PR commits.

- **Matrix (8 → 6 jobs):** Dropped Numba to Python 3.13 only — Numba compiles through its own LLVM backend, so running it on both versions mostly duplicates coverage. Python-only modes still run on 3.10 and 3.13. Saves ~240 minutes per run.

- **Refactor:** Collapsed the four conditional run steps into one matrix-driven step (matches the Docker workflow), moved `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` to workflow-level, and added an mpi4py import check after install so a silent wheel override from `[dev]` extras would show up immediately in the logs.

***Docker workflow:*** Switched to `docker/build-push-action@v5`, refactoring

Net effect: roughly half the runner minutes per PR, faster Docker builds via cache reuse, and published images are now traceable to a specific commit.

## Developer Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,12 +3,6 @@
 # =============================================================================
 # Builds the CPU Docker image, runs the full test suite inside it,
 # and publishes to GHCR on pushes to main/dev (i.e., after PR merge).
-#
-# Publish trigger logic:
-#   - PRs        → build + test only (no publish)
-#   - Push to dev/main → build + test + publish
-#   When a PR is merged, GitHub fires a 'push' event on the target branch,
-#   which naturally triggers the publish path.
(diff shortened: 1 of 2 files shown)


ilhamv:
This looks great and clearer! Thanks, Melek.
@melekderman - Let's just test only on Python 3.13, removing testing on 3.10.

melekderman:
@ilhamv,

Sounds good! Then I have removed tests on Python 3.10.

PR #440, opened 2026-06-03
Add GPU Functionality to the `dev` Branch

### LINKED ISSUE #479: Update GPU Support for Compatibility with Refactored Codebase
## Background

As MC/DC’s is refactored, the GPU support layer needs to be updated so it remains compatible with the new structure.

This effort should include input from transport algorithm developers so that GPU-facing interfaces are balanced between practical usability and performance requirements.

## Goal

Update GPU support to work cleanly with the refactored `dev` architecture, while providing an interface that is usable for transport algorithm development and still suitable for performant GPU execution.

## Scope

This work should include:

- Reviewing the current GPU support layer against the refactored `dev` architecture.
- Identifying compatibility gaps introduced by the refactor.
- Updating GPU data structures, preparation pathways, execution hooks, and runtime interfaces as needed.
- Gathering input from transport algorithm developers on the GPU-facing interface.
- Clarifying which parts of the interface should prioritize usability, extensibility, or performance.
- Preserving or improving current GPU performance where practical.
- Updating tests, examples, and developer documentation affected by the GPU changes.

## Acceptance Criteria

- GPU support is compatible with the refactored `dev` architecture.
- Transport algorithm developers can use the GPU interface without relying on undocumented internal behavior.
- GPU-facing APIs are documented clearly enough for framework and algorithm developers.
- The interface balances practical usability with performance-sensitive implementation needs.
- Existing GPU workflows either continue to work or have documented migration paths.
- GPU tests pass, or known limitations are documented.
- Performance-sensitive changes are benchmarked or otherwise assessed.

## Notes

This should be treated as both a compatibility update and an interface-design effort. The GPU layer should not merely be made to run again; it should be reviewed with transport algorithm developers so the resulting design supports future algorithm work without hiding critical performance constraints.

braxtoncuneo:
## Summary of changes

This PR adds GPU execution to the dev branch. This was accomplished by re-implementing some functionality in a manner compatible with execution through `numba-hip` and `numba-cuda`. This includes changes to how gpu state objects are managed, the introduction of helper intrinsics/functions to handle returning arrays, and replacement of builtin functions that are not supported on one or more gpu plaltforms (e.g. powers for complex numbers, numpy polynomial solvers).

## Types of changes
**New feature: Helper functions/intrinsics for returning arrays on GPU**

As has been mentioned in previous issues/PRs, `numba-hip` and `numba-cuda` do not permit returning arrays from functions. In brief, this is because memory management on GPU is hard, and the simplest way to create an array as a local variable is to allocate it on the stack. Once the array that created a local array returns, that array no longer "exists" in a way that it can be safely used.

This is a problem, since the helper functions provided by `numba_objects_generator.py` return references to and slices of arrays in the mcdc global data structure. We'd like to keep those helper functions, considering how much convenience and quality of life they provide to developers. Also, since these global data structures have a lifetime which encompasses the lifetime of all transport logic (i.e. the structures are in memory and safe to use whenever any transport logic is run) it should be safe to ignore this array-returning restriction for these helper functions.

For this specific use case, *when we know for certain that an array is safe to use after a function returns*, this PR introduces the `array_return` decorator and the `array_result` function, which enables the return of array references/slices like so:

```python
# The decorator must be supplied the type of element stored
# by the arrays that are returned by the decorated function.
# 
# An arbitrary number of parameters may be accepted by an
# array-returning function, as long as that number is fixed
# for a given array returning function.
@array_return(nb.types.float64)
def example_function(param_1,param_2,param_3):
    # ...array slicing logic...

    # Returned arrays must always be wrapped in the
    # array_result function.
    return array_result(example_result_array_here)
```

This works by smuggling a pointer to the array's data and the size of the array out as a tuple, then reconstructing the array based off of the returned tuple in the calling function.

In normal python execution, the `array_result` function simply returns its input. However, there exists an overload of `array_result` which will handle the conversion to tuple in compiled contexts.

The `array_return` decorator registers an intrinsic overload of the decorated function. This intrinsic calls the decorated function and creates a new array based on the returned tuple. Because intrinsics are inlined directly into their calling function, they aren't their own function with a separate stack frame and so are not rejected by the array-returning guards of `numba-hip` or `numba-cuda`.

**New feature: GPU-compatible quartic solver and complex power calculation**

- **Good news:** The torus has been added as a new geometric primitive to MC/DC!
- **Bad news:** Solving torus/line intersections analytically requires solving quartic functions.
- **Good news:** `numpy.roots` solves polynomials and is implemented by `numba` in compiled contexts!
- **Bad news:** `numba-hip` and `numba-cuda` *don't* implement `numpy.roots` - likely because the general case does not have analytic solutions and requires the use of arbitrarily large matrixes (determined by the degree of the polynomial).
- **Good news:** Quartic functions *do* have an analytic solution!
- **Bad news:** Quartic functions require the calculation of real-valued powers for complex numbers, and  `numba-hip`  *doesn't* implement the intrinsic for powers of complex numbers.
- **Good news:** Integer powers of complex numbers can be calculated through iterative multiplication, complex square roots can be solved with square roots of real numbers, and fractional powers of complex numbers can be calculated using de Moivre's Formula!

```python
@njit()
def modulus(x):
    return math.sqrt(x.real**2 + x.imag**2)


@njit()
def sqrt(x):
    return math.sqrt(x.real) * (x + x.real) / modulus(x + x.real)


@njit()
def power(x, n):
    result = 1
    for i in range(n):
        result = result * x
    return result


@njit()
def nth_root(x, n, index):
    # First, convert to polar form
    r = modulus(x)
    a = math.atan2(x.imag, x.real)

    # Apply de Moivre's Formula
    root_modulus = math.pow(r, 1.0 / n)
    root_argument = (a + 2 * math.pi * index) / n

    # ...then convert back to rectangular form
    real = root_modulus * math.cos(root_argument)
    imag = root_modulus * math.sin(root_argument)
    return complex(real, imag)


@njit()
def principal_nth_root(x, n):
    return nth_root(x, n, 0)
```

**New feature: GPU execution**

**New feature:** Added logic to `test/regression/conftest.py` to support regression/units tests with GPU

**Bug fix:** updated references to reflect the new names and module locations of Harmonize API, as imported by MC/DC

**Bug fix:** updated logic switched by gpu-related settings to check values in `settings` based upon the constants in `mcdc.constant`

**Bug fix:** removed redundant call to `clear_flags` in `code_factory/gpu/transport/simulation.py`

**Bug fix:** ensured that the original value of atomically-modified fields is taken from the return of the atomic operation, rather than from a separate read. This is important, because only the return value of the atomic operation itself can be trusted with providing the original value immediately before the operation.

**Bug fix:** updated `add_bank_size` and its uses so that they follow the interface of atomic operations, returning the trusted value. Using a separate `get_bank_size` after the fact has the same problem as a separate read, and so the original value must be provided by the logic of `add_bank_size` itself.

**Bug fix:** modified the reading of particles from banks so that they use the trusted values returned by `add_bank_size`, which requires the movement to occur after the addition itself has occurred.

**Bug fix:** added early return to progress-printing code in `mcdc/transport/simulation.py` to avoid divide-by-zero when `mpi_work_size==0`

**Bug fix:** updated the cpu-side `atomic_add` to match the correct atomic operation interface

**Refactor:** Added gpu platform detection to `config.py`, providing global flags to indicate whether CUDA/ROCM is available

**Bug fix:** Moved the call to `setup_gpu_program` in `main` so that it occurs after types have been defined. Without this ordering, the function would attempt to compile functions without proper type information.

**Bug fix:** Updated `test/unit/test_numba_layers_generator.py` to reflect the new array-returning functionality provided for `mcdc_get`/`mcdc_set`.

## Developer Checklist

## Associated Issues and PRs

- related to #


braxtoncuneo:
It should be noted that the `slab_isobeam_td_census` example was removed from the regression tests, which failed close-to-identically on both `python`/`numba`-mode and both with `cpu` or `gpu` as the target.

After consultation with @ilhamv , this seemed like the best option to get gpu functionality merged in a timely manner.

mcdc/code_factory/gpu/transport/simulation.py
```diff
@@ -39,10 +39,10 @@ def source_loop(seed, simulation, data):
 
         # Store the global state to the GPU
```
braxtoncuneo, line 43: The place where runtime API entry points are loaded in from Harmonize changed. The name of the module and the name of the globals they are loaded into needed to be updated.

mcdc/code_factory/gpu/transport/util.py
```diff
@@ -5,9 +5,28 @@
...
+
+
+@nb.extending.overload(atomic_add, target="cpu")
+def overload_atomic_add_cpu(array, idx, value):
+    def impl(array, idx, value):
+        result = array[idx]
+        array[idx] += value
+        return result
+
+    return impl
```
braxtoncuneo, line 29: The `atomic_add` function is called both on cpu and on gpu when running MC/DC for gpu. An overload needed to be made so the function can safely be called on either platform.

mcdc/code_factory/gpu/program_builder.py
```diff
@@ -252,6 +347,11 @@ def step(program: nb.uintp, particle_input: particle_gpu):
         particle_container = util.local_array(1, type_.particle)
         particle_container[0] = particle_input
         particle = particle_container[0]
+        particle["alive"] = True
+        particle["material_ID"] = -1
+        particle["cell_ID"] = -1
+        particle["surface_ID"] = -1
+        particle["event"] = -1
```
braxtoncuneo, line 262: Particle initialization logic for when particles are loaded from banks on GPU.

mcdc/code_factory/gpu/program_builder.py
```diff
@@ -297,6 +397,30 @@ def step(program: nb.uintp, particle_input: particle_gpu):
...
+    loop_source = gpu_loop_source
+    #
+    # Overwrite function
+    for impl in target_rosters["cpu"].values():
+        overwrite_func(impl, impl)
+    """
+
+    alloc_managed_bytes = harmonize.alloc_managed_bytes
+    alloc_device_bytes = harmonize.alloc_device_bytes
+
```
braxtoncuneo, line 318: Additional runtime API entry points that were added.

mcdc/code_factory/gpu/program_builder.py
```diff
@@ -342,31 +465,6 @@ def teardown_gpu_program(simulation):
...
-def create_mcdc_container(dtype):
-    if config.gpu_state_storage == "managed":
-        mcdc_ptr = harmonize.alloc_managed_bytes(dtype.itemsize)
-    else:
-        mcdc_ptr = harmonize.alloc_device_bytes(dtype.itemsize)
-    mcdc_uint = cast_voidptr_to_uintp(mcdc_ptr)
-    mcdc_container = nb.carray(mcdc_ptr, (1,), dtype)
-    return mcdc_container, mcdc_uint
-
-
```
braxtoncuneo, line 369: These were redundant definitions of functions defined elsewhere, and were no longer used.

mcdc/code_factory/array_return.py
```diff
@@ -0,0 +1,288 @@
...
+
+# A function decorated with `array_return` may return an array by passing
+# it through the `array_result` function and returning the output
+def array_return(sig, ndim=1):
+    def array_return_true_decorator(fn):
+        array_return_typing(fn, sig, ndim)
+        array_return_lowering(fn, sig, ndim)
+        return fn
+
+    return array_return_true_decorator
```
braxtoncuneo, line 288: This is where the array-returning utility logic is defined. This is probably the most complex addition.

This code will probably be difficult for some maintainers to understand/modify, since it interacts so much with Numba's extension systems. It should not need to be modified regularly, so this would likely not be a problem.

test/unit/test_numba_layers_generator.py
```diff
@@ -129,8 +130,8 @@ def test_scalar_integer_getters_cast_values_from_data():
 
 def test_float_and_bulk_getters_remain_zero_copy_views():
     assert "return data[offset + index]" in _accessor_1d_element("example", "values")
-    assert "return data[start:end]" in _accessor_1d_all(
-        "example", "values", "values_length"
+    assert "return array_result(data[start:end])" in _accessor_1d_all(
+        "example", "values", "values_length", nb.types.float64
```
braxtoncuneo, line 134: This update reflects the new array-returning logic provided in this PR, allowing this unit test to pass on GPU.

test/unit/test_annotation_shape.py
```diff
@@ -66,8 +67,8 @@ def test_stringified_annotation_rejects_incorrect_offset_shape(capsys):
 
 
 def test_generated_accessor_resolves_dimension_offset():
-    all_source = _accessor_1d_all("mgxs", "energy", "G+1")
-    last_source = _accessor_1d_last("mgxs", "energy", "N - 2")
+    all_source = _accessor_1d_all("mgxs", "energy", "G+1", nb.types.float64)
+    last_source = _accessor_1d_last("mgxs", "energy", "N - 2", nb.types.float64)
```
braxtoncuneo, line 71: The element type of the array being returned by the resulting function is provided to match the revised interface needed by the new array-returning logic.

test/regression/conftest.py
```diff
@@ -182,7 +192,8 @@ def compare_tallies(output, answer, target, errors):
                 continue
             name_score = f"{name_tally}/{score}"
             for result in answer[name_score].keys():
-                if "uq_var" in result and target == "gpu":
+                should_pass = any([x in result for x in gpu_pass_list])
+                if should_pass and target == "gpu":
                     continue
```
braxtoncuneo, line 197: Due to how GPU execution batches samples, it is practically impossible to calculate the same standard deviation that CPU execution produces. This removes the need to match standard deviation output in regression testing.

test/regression/conftest.py
```diff
@@ -136,11 +136,20 @@ def build_command(config):
     target = config.getoption("--target")
     mpiexec = config.getoption("--mpiexec")
     srun = config.getoption("--srun")
+
+    state = ""
+    if target == "gpu":
+        state = "--gpu_state_storage=managed"
+        mode = "numba"
+
```
braxtoncuneo, line 144: Currently, the default "separate" storage option does not work. Also, gpu execution requires `mode=numba`. This logic ensures that settings are updated to support gpu execution.

test/regression/conftest.py
```diff
@@ -136,11 +136,20 @@ def build_command(config):
...
     command = [
         sys.executable,
         "input.py",
+        f"--clear_cache",
+        f"--caching",
```
braxtoncuneo, line 149: Caching can break execution if you switch between cpu and gpu-based regression testing. These additional flags fix this.

mcdc/main.py
```diff
@@ -170,6 +170,15 @@ def prepare(simulationPy: Simulation):
...
 
+    # ==================================================================================
+    # Setup GPU-Related Data Structures, if Necessary
+    # ==================================================================================
+
+    if config.target == "gpu":
+        from mcdc.code_factory.gpu.program_builder import setup_gpu_program
+
+        setup_gpu_program(simulation_container, data)
+
```
braxtoncuneo: Previously, `setup_gpu_program`'s call was placed before all types were defined, which means that the compilation performed by `setup_gpu_program` is missing crucial type information. This PR fixes that ordering.

mcdc/config.py
```diff
@@ -114,6 +114,28 @@ def _build_parser() -> argparse.ArgumentParser:
...
+    try:
+        import numba.cuda as cuda
+
+        CUDA_AVAILABLE = True
+    except:
+        CUDA_AVAILABLE = False
+else:
+    CUDA_AVAILABLE = False
+
+
```
braxtoncuneo, line 138: The array-returning functionality requires knowledge about which gpu platform is available. To support use in other potential future additions, the platform detection logic has been added to `config.py`, rather than being placed directly in the array-returning logic.

mcdc/transport/util.py
```diff
@@ -153,14 +154,23 @@ def log_interpolation(x, x1, x2, y1, y2):
...
 def access_simulation(program):
     return program
+
+
+@nb.extending.overload(access_simulation, target="cpu")
+def access_simulation_cpu_overload(program):
+    def impl(program):
+        return program
+
+    return impl
```
braxtoncuneo, line 176: Since `for_cpu` and `for_gpu` have been removed, and the state accessor functions perform fundamentally different operations on different platforms, they need to be overloaded directly.

mcdc/transport/util.py
```diff
@@ -153,14 +154,23 @@ def log_interpolation(x, x1, x2, y1, y2):
 
 @njit
 def atomic_add(array, idx, value):
+    result = array[idx]
     array[idx] += value
+    return result
```
braxtoncuneo, line 159: Atomic operations must always report the original value.

mcdc/code_factory/jit.py
```diff
@@ -0,0 +1,54 @@
...
+        njit_fn = nb.njit(args[0])
+        return wrap_with_check(fn, njit_fn)
+
+    else:
+        if not ARG_CHECK:
+            return nb.njit(*args, **kwargs)
+
+        def wrapper(fn):
+            njit_fn = nb.njit(*args, **kwargs)(fn)
+            return wrap_with_check(fn, njit_fn)
```
braxtoncuneo: An alternative to `njit` for injecting wrapper logic around compiled code.

mcdc/code_factory/numba_layers_generator.py
```diff
@@ -1259,7 +1289,7 @@ def _accessor_2d_vector(object_name, attribute_name, stride, setter=False):
     if setter:
         text += f"    data[start:end] = value\n\n\n"
     else:
-        text += f"    return data[start:end]\n\n\n"
+        text += f"    return array_result(data[start:end])\n\n\n"
     return text
```
braxtoncuneo, line 1284: These changes add the array-returning helper functions to the code generated in `mcdc_get` and `mcdc_set`, allowing code to use array-returning accessors in gpu execution.

braxtoncuneo:
These latest changes remove dead code, and so should not affect regression/unit tests.

test/regression/conftest.py
ilhamv: I see. If all the regression tests are made batch-wise, they should be reproducible, right?

ilhamv:
Thanks, @braxtoncuneo! I'll work on the review ASAP. And thanks for highlighting some of the changes in separate conversations.

mcdc/transport/util.py
ilhamv: Why is that? Is it because the GPU mode needs to know the original value after each atomic_add call?

ilhamv:
Dedicated pages in `docs` would be needed to reflect the updated GPU support. We will do that in a follow-up PR.

ilhamv:
We would need to keep track of overheads that appear only in pure Python mode, as it would be important in interpreting runtime results when we compare Python VS Numba-CPU modes. We can do this when we work on the docs in the follow-up PR.

mcdc/transport/util.py
braxtoncuneo: Basically, yes. More precisely, concurrent code needs to know what the original value is in many important cases. In a potential future where cpu-side thread parallelism is in use, this would also be necessary.

In a nutshell, atomic operations represent an operation performed on a value in memory which is guaranteed not to be interfered with by any other concurrent operation. For example, if we were to perform a conventional read/modify/write operation with tallies

```python
old_total = tallies[x,y,z]
new_total = old_total + weight
tallies[x,y,z] = new_total
```
we could end up in a situation where threads mess with each other by interleaving parts of this accumulation
```
thread A: old_total = tallies[x,y,z] // 100
thread A: new_total = old_total + weight // 100.5
thread B: old_total = tallies[x,y,z] // 100
thread A: tallies[x,y,z] = new_total // 100.5
thread B: new_total = old_total + weight // 101
thread B: tallies[x,y,z] = new_total // 101
```
It's tempting to think something like this would solve the problem:
```python
tallies[x,y,z] += weight
```
but these compound operations end up getting compiled into separate read/modify/write operations akin to the first code snippet.

Atomic operations fill this gap by providing a slightly more expensive instruction that locks the target bytes in memory for exclusive use by the thread and performs the read/modify/write during that lock period. Other threads that try to do atomic operations on that region of memory at that exact time will be forced to wait their turn.

There are times where we just want to add to a value and don't necessarily care what the old value was -- for example, accumulating into tallies. In that case, we just care that the no other concurrent operations interfered with the addition. A common example of this problem is with accumulating tallies.

*However*, atomic operations are also commonly used to track state transitions in data structures, where the precise state transition caused by the atomic operation has a direct impact on what must happen afterwards.

A pertinent example of this is storing into a particle bank: we want to reserve a space in the particle bank to store a particle. The bank grows upwards from the zero-index element, so we will want to add to the size of the bank.

If we do a normal addition at this part of the code, some of the incrementations made by threads would likely be lost - leading to a final bank size that is smaller than it should be. If we use an atomic addition, all incrementations are guaranteed to be counted in the final total *BUT* there is no guarantee that any separate read made by a thread will reflect the state that was most recently induced by an atomic operation it performed. In other words, we would have no guarantee that each additional particle in the bank would receive a unique index, because each thread has no guarantee of observing a unique bank size with each normal read. The only way we could reasonably do so is by knowing the exact state transition caused by the corresponding increment. This is why atomic operations themselves must be charged with returning information about the state of the target variable: it is the only thing that is capable of retrieving state information during its brief lock period.

A common follow-up question is *why do atomic operations return the original state*? This is because not all binary operations are reversible based upon one input and the output. For example, if you did a bitwise atomic `AND` between the mystery byte `XXXXXXXX` and `00001111` and knew the result was `00000000`, you would have perfect information about the last four bits in the mystery value but would not be able to recover any bits in the first four bits. However, if the operation is deterministic and you know both inputs, the output can be determined readily - allowing a program to know both the starting and ending state of the transition.

test/regression/conftest.py
braxtoncuneo: They certainly should be. We're reproducing the tallies quite closely in output, a difference on the order of about 1e-17. Unless particles managed to sneak from one batch into another, the tally totals for each individual batch should also match that closely.

mcdc/transport/util.py
ilhamv: I see. Many thanks for the thoughtful explanation, Braxton!

mcdc/transport/simulation.py
```diff
@@ -256,6 +256,9 @@ def source_closeout(simulation, idx_work, N_prog, data):
             tally_module.closeout.accumulate(simulation, data)
 
     # Progress printout
+    if simulation["mpi_work_size"] == 0:
+        return
+
```
ilhamv: @braxtoncuneo 
Is it possible to have mpi_work_size = 0?

ilhamv:
@braxtoncuneo 

`setup_gpu_program` is called twice, first in `generate_numba_layers()` then in `prepare()`. Is that on purpose?

ilhamv:
@braxtoncuneo 
I think I fixed the bug in `slab_isobeam_td_census`. The issue was in using the returned value of add_bank_size when we promote a future particle to a census particle.

ilhamv:
@braxtoncuneo 
If now mcdc_get supports Numba-compatible array returns, should we "release" all the suppressed instances of "Above is equivalent to:"?

mcdc/transport/geometry/interface.py
```diff
@@ -369,7 +369,7 @@ def _check_cell(particle_container, speed, cell, simulation, data):
         return True
 
     # Create local value array
-    value = util.local_array(literals.rpn_evaluation_buffer_size(), np.bool_)
+    value = util.local_array(100, np.bool_)
```
ilhamv: @braxtoncuneo 
Does the rpn_evaluation_buffer_size still work? If it does, we should keep it.

mcdc/code_factory/jit.py
ilhamv: @braxtoncuneo 
Is this exclusively used for the GPU mode?

ilhamv:
@braxtoncuneo 
Note that I organize the root solver, renaming `geometry/root_solve.py` into `geometry/surface/torus_root_solver.py`, as it is currently used exclusively for torus. 

I can see that there is potential for generalizing it to be a utility module. However, generalizing it would need to make the functions tangent-safe, like those currently used in the quadratic surfaces, not to mention the sacrificed performance from moving away from the specialized solvers. Nevertheless, due to the potential improved maintainability, we keep the generalization option open in the future!

mcdc/transport/simulation.py
braxtoncuneo: It is certainly possible to have `mpi_work_size == 0`, though I agree it should not be the case for this function.

In `manage_particle_banks`, `mpi.distribute_work` is called with the size of the source bank as the `N_work` parameter:
- `manage_particle_banks` sets the size of the source bank as the size of the census bank (as part of transferring particles from the census bank to the source bank)
- `distribute_work` sets `simulation["mpi_work_size"]` based on `N_work`
- If the census bank from the previous source loop has no particles, the source bank is set as empty, meaning the `N_work` parameter is set as zero, meaning `simulation["mpi_work_size"]`  is set to zero

This divide-by-zero is avoided on cpu because the cpu `source_loop` runs `source_closeout` once per particle. The current version of the gpu `source_loop` runs `source_closeout` once per source loop.

You can find an example of this happening with the `inf_shem361_td_census` regression test.

I'll move this if statement from `source_closeout` to the top of the gpu `source_loop`, that way the additional overhead of starting/stopping the loop is avoided.

braxtoncuneo:
> @braxtoncuneo If now mcdc_get supports Numba-compatible array returns, should we "release" all the suppressed instances of "Above is equivalent to:"?

You're right. I'll get those switched back to the `mcdc_get` equivalents.

mcdc/code_factory/jit.py
braxtoncuneo: I've used this to detect when optional/record values are passed as arguments when they are not supposed to be. Since it introduces overhead to compilation, it would not substitute the regular `numba.njit` unless necessary.

Upon review, it looks like the mechanism to substitute this njit for `numba.njit` didn't survive the recent rebase on the new repo's `dev` branch. For now, since it doesn't affect the rest of the code, I'll just move the file to a separate branch and add in the parameter type diagnostics as a separate PR.

braxtoncuneo: @ilhamv  The latest set of commits added to this PR removes `mcdc/code_factory/jit.py`

mcdc/transport/simulation.py
braxtoncuneo: @ilhamv The latest set of commits added to this PR moves the early return from  `source_closeout` to `source_loop`.

braxtoncuneo:
> @braxtoncuneo
> 
> `setup_gpu_program` is called twice, first in `generate_numba_layers()` then in `prepare()`. Is that on purpose?

No. In fact, it looks like some of the changes between the CEMeNT `dev` and the MCDC project dev already fixed the ordering problem for `setup_gpu_program` and defining types. This latest batch of commits removes the additional call.

> @braxtoncuneo I think I fixed the bug in `slab_isobeam_td_census`. The issue was in using the returned value of add_bank_size when we promote a future particle to a census particle.

I have verified that it passes for gpu on tuolumne. :+1: 

> @braxtoncuneo If now mcdc_get supports Numba-compatible array returns, should we "release" all the suppressed instances of "Above is equivalent to:"?

The latest batch of commits revises these accesses back to the equivalent `mcdc_get` uses.

mcdc/transport/geometry/interface.py
braxtoncuneo: It works for gpu on tuolumne, as of the latest commit. This change is now removed.

ilhamv (approved):

PR #471, opened 2026-07-11
fix: 2D-vector setter writes nothing (`-` instead of `=`)

steps-re:
`_accessor_2d_vector` in `mcdc/code_factory/numba_objects_generator.py` generates its setter with:

```python
text += f"    data[start:end] - value\n\n\n"
```

That emits `data[start:end] - value` — a subtraction whose result is thrown away — instead of `data[start:end] = value`. Every sibling accessor generator (`_accessor_1d`, `_accessor_2d`, `_accessor_3d_element`, ...) correctly emits `= value`, so this one is the odd one out and every generated `*_vector` setter silently writes nothing.

This fixes the generator template and the checked-in generated modules where the broken setter already landed:
- `mcdc/mcdc_set/table_data.py`
- `mcdc/mcdc_set/multigroup_material.py` (4 setters)
- `mcdc/mcdc_set/source.py` (2)
- `mcdc/mcdc_set/neutron_inelastic_scattering_reaction.py`
- `mcdc/mcdc_set/surface.py` (2)

If those modules are meant to be regenerated from the template rather than hand-edited, let me know and I'll drop the generated-file changes and keep only the template fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/2.0.0/),
 
 ### Fixed
 
+- Fix 2D-vector setter writes nothing (- instead of =) from [@steps-re]
+
 ### Security
 
 ## [0.14.1] - 2026-07-04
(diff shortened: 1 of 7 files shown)


ilhamv:
Thanks, @steps-re!

ilhamv (approved):
Thanks, @steps-re!

This PR will be part of patch release v0.14.2, along with #470.

PR #530, opened 2026-09-08
Update numba support rebuild script and rebuild per GPU patch

ilhamv:
## Summary of changes

- Filter out empty accessors from creation
- Remove redundant imports


braxtoncuneo (approved):
This PR passes all regression tests with `--mode=GPU` on tuolumne with `python=3.11.5`.

Given `.github/workflows/check_numba_support.yml` uses `python=3.13` and still passes on CPU in the CI, it looks like the source of the issue solved by #534 may be specific to GPU execution. When GPU regression tests get added into CI, it may be good to include a GPU numba support check to guard against any GPU-specific discrepancies.

Nonetheless, looks good. :tada:

PR #531, opened 2026-09-09
Add standard performance metrics and optional tally output

ilhamv:
## Summary of changes

Add `performance/` in the HDF5 output with total runtime, history count, MPI rank count, and effective variance.
Compute effective variance during compiled tally closeout.
Replace `--runtime_output` with `--no-tally_output` to skip saving tally results.
These quantities and options are useful for the performance test suite, where we are mainly assessing performance metrics rather than the actual tally outputs.

## Types of changes

- New feature (non-breaking change which adds functionality) 
- Organization and beautification (changes which improve readability and/or accessibility) 

## Developer Checklist

- [x] I have read the [contributing guide](https://mcdc.readthedocs.io/en/dev/contributing/index.html).
- [x] My code follows the [code style](https://mcdc.readthedocs.io/en/dev/contributing/index.html#code-styling) of this project.
- [x] I have updated the documentation as needed. 
- [x] I have updated `CHANGELOG.md` as needed. 
- [x] I have added or updated tests to cover my changes. 


massimolarsen (commented):
The PR looks good overall. I also took at look at the TODO in _finalize() while I was looking at the PR

``` python
# Check for round-off error (TODO: Check why this is needed.)
if abs(radicand) < 1e-16:
    data[offset_sum_square + i] = 0.0
```

I think this check was probably put in place to fix small negative values of the radicand caused by roundoff (which would hit the sqrt -> NaN). However, it also discards valid small positive variances. Floating point precision is relative to the magnitude of the values, so being smaller than `1e-16` does not necessarily make a value unreliable.

For example, if we have two scores `[1, 3]`, the radicand is `1` and the squared relative std error is `0.25`. Scaling those scores to `[1e-9, 3e-9]` gives a radicand of `1e-18`, but the squared relative std error should still be `0.25`. Instead, this check sets the std error to zero, so the squared relative std error which should be `0.25` becomes `0.0`. And then this goes on to affect the new variance metrics.

I think we could mimic finalize_census and use `max` to clamp the negative values and leave the positive ones alone:

```python
radicand = max(radicand, 0.0)
data[offset_sum_square + i] = math.sqrt(radicand)
```

ilhamv:
Good catch, @massimolarsen — I'll change it to that. Thanks!

ilhamv:
@massimolarsen - I also made minor changes for readability and consistency between finalize and finalize_census.

ilhamv:
Interesting, the change exposes the cases where stdev is incorrectly zeroed out due to the aggressive round-off error check. Since it's a very well-isolated fix, I'll just go ahead and update the regression test answer.

ilhamv:
OK, now one regtest is still failing, but only in Python mode...

Looks like there is a round-off error issue related to different Python versions (3.13.3 VS 3.13.5) or platforms (macOS VS Linux) used...

ilhamv:
I'm working on a more robust way to accumulate the mean and the second moment. New update coming soon.

ilhamv:
@massimolarsen  - The main change I made is replacing the raw-moment calculation with Welford’s online variance algorithm:
```python
n += 1
delta = x - mean
mean += delta / n
M2 += delta * (x - mean)
```
This avoids the problematic subtraction of two large, nearly equal moments.

Other than that, the changes are organizational, tailored around the main change.

massimolarsen (approved):
Good call with the Welford's algorithm, I probably should have done that to begin with. Everything else looks good to me, and the tests are passing.

PR #533, opened 2026-09-10
Use buffer for GPU mode during MPI reduce

ilhamv:
Preserve in-place MPI tally reductions for CPU and separate GPU storage. Use a reusable host receive buffer for other GPU storage models to address the reported multi-rank GPU bus error.


braxtoncuneo (approved):
All regression tests pass with `--mode=gpu` on tuolumne.

LGTM

PR #534, opened 2026-09-11
Support named NumPy NDArray aliases

ilhamv:
Normalize named NDArray[scalar] aliases before decoding array annotations, preserving compatibility with the expanded numpy.ndarray[shape, dtype] form. This fixes annotation unpacking failures and recognition of unannotated array fields for certain Python versions.


ilhamv:
Approved per https://github.com/mcdc-project/mcdc/pull/530#pullrequestreview-5185064573

PR #528, opened 2026-09-08
optimize tally moments storage

ilhamv:
## Summary of changes

This PR optimizes tally storage by only allocating tally moments to the master rank. The tally moments are still allocated to the non-master rank if we run a one-batch simulation only.

## Types of changes

- Optimization

## Developer Checklist

- [x] I have read the [contributing guide](https://mcdc.readthedocs.io/en/dev/contributing/index.html).
- [x] My code follows the [code style](https://mcdc.readthedocs.io/en/dev/contributing/index.html#code-styling) of this project.
- [x] I have updated the documentation as needed. 
- [x] I have updated `CHANGELOG.md` as needed. 
- [x] I have added or updated tests to cover my changes. 
- [x] All new and existing tests pass

## Associated Issues and PRs

- closes #
- related to #

## Associated Developers

- @

PR #535, opened 2026-09-11
Reintroduce Original Closeout Reduction for Potentially Misaligned Data Buffers

braxtoncuneo:
## Summary of changes

Reintroduces the original closeout reduction removed by #528, but only for cases where MCDC's `data` array is stored in `united` storage.

While the memory provided by `hipMalloc` is suitably aligned for operations provided by the ROCm runtime, it may not be suitably aligned for all native functions used by MCDC. In particular, the implementation of in-place reduction provided by `cray-mpich` (e.g. on tuolumne) appears to apply an unaligned vector load/store operation which results in a bus error.

Until a suitable alignment fix for united storage is established, this PR will allow MCDC to fall back on the original reduction logic when necessary.

In contrast, `managed` storage does not suffer from bus errors when running the in-place reduction, likely due to the CPU and GPU using separate page tables, so it will continue to use in-place reduction.

## Types of changes

- Bug fix: Stops crashing on tuolumne when using both MPI and `--gpu_state_storage=united`

## Developer Checklist

- [X] I have read the [contributing guide](https://mcdc.readthedocs.io/en/dev/contributing/index.html).
- [X] My code follows the [code style](https://mcdc.readthedocs.io/en/dev/contributing/index.html#code-styling) of this project.
- [ ] All new and existing tests pass

While this does fix a problem, the problem in question is not from a previous release, hence no updates to `CHANGELOG.md`.

## Associated Issues and PRs
- related to #528

## Associated Developers
- @ilhamv


ilhamv:
Thanks, @braxtoncuneo — Does this PR solve the same problem as in #533?
Or, perhaps this fixes an issue that the former one leaves unfixed?

Nevertheless, let's work on merging #530 first, as that one deals with the unsynchronized numba supports (numba_types, mcdc_get, mcdc_set).

braxtoncuneo:
Based on testing, it appears that execution with `--gpu_state_storage=managed` may not require this workaround.

Still, this PR does restore the original behavior, and we could test on additional platforms before permitting in-place reductions on arrays stored in managed allocations. It's likely best we hold off on that until after the upcoming version release.

